PR #3, opened 2023-09-20
Made pkg private, refactored metadata_library.go

dhirajsb:
## Description
Added go-enum local install instructions to README.md.
Made pkg code private to avoid exporting server internals outside the project. 
Refactored metadata_library to make it reusable in other tools to come like GraphQL schema generator. 
Removed generated gRPC go model in .gitignore to support checking if gen/grpc was run for a PR to avoid breaking code from any changes to proto files. 

## How Has This Been Tested?

Tested build works and ran `make run/client` against `make run/server` to verify client and server work. 

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.


cmd/migrate.go
```diff
@@ -96,82 +91,51 @@ func migrateDatabase(dbConn *gorm.DB) error {
...
+				ExternalId:  et.ExternalId,
+				Properties:  library.ToProtoProperties(et.Properties),
+			})
+		}
+		response, err := grpcServer.PutTypes(context.Background(), &typesRequest)
 		if err != nil {
-			return fmt.Errorf("failed to read library directory %s: %w", abs, err)
+			return fmt.Errorf("failed to add library from file %s: %w", path, err)
 		}
+		glog.Infof("created/updated %d artifacts, %d contexts and %d execution types from library file %s",
```
tarilabs:
```suggestion
		glog.Infof("created/updated %d ArtifactTypes, %d ContextTypes and %d ExecutionTypes from library file %s",
```

tarilabs (dismissed):
I'd recommend to keep scope of PR small, refactoring locations vs refactoring functions _possibly_ separate as it's now always easy to follow on github 😅 

very minor comment on logging using domain object precise naming, totally optional

internal/model/library/metadata_library.go
```diff
@@ -1,7 +1,14 @@
...
+	"github.com/golang/glog"
+	"github.com/opendatahub-io/model-registry/internal/ml_metadata/proto"
+	"gopkg.in/yaml.v3"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
 )
 
 //go:generate go-enum -type=PropertyType
```
lampajr, line 13: Running `make build` I hit this issue here:
```
❯ make build
go generate ./...
internal/model/library/metadata_library.go:14: running "go-enum": exec: "go-enum": executable file not found in $PATH
make: *** [Makefile:34: gen] Error 1
```

Issue that I solved running `go install github.com/searKing/golang/tools/go-enum@v1.2.97`, is that expected? or am I missing something?

If that is expected may I suggest to add it in the list of Go tools pre-requisites in the README?

antisaling: As this was already here before, we should probably fix it separately from this PR. It seems it really should be added as a prereq.

lampajr: yeah it was already there before but it was ignored IIUC, I think it started failing because of https://github.com/opendatahub-io/model-registry/pull/3/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R34 that is why I would include the new prereq here. Anyway not big deal if we want to fix it here or in a separate PR, just wanted to be sure it was needed/expected :smile:

antisaling (changes requested):
We should stick to git's imperative style in commit messages. Also, can you make the commit message shorter as it's bluntly cut in GH UI? Something like: "Make pkg code private", then elaborate on the change (i.e. "As it's not meant to be exported...") on the description.

I'm not sure if it's possible to separate the library loading change in another commit, but then describe that change (what and why, not how) in the commit description as well, please.

I still have to test it, will follow up later.

internal/model/library/metadata_library.go
antisaling: I see, nice catch! I don't mind including it on this PR, probably could be inserted as a first commit or something.

dhirajsb: Right, need go-enum install in the README as well. I'll add that as a separate commit.

dhirajsb: Added instructions for installing go-enum.

dhirajsb:
Fixed commit messages and added descriptions. It's an old habit from short single line commit messages from old Apache projects. 
I'm all for standards and process, but let's make sure it doesn't consume too much of our time and become a chore in itself. :+1:

tarilabs (approved):
seems to me include previous optional comments, and other comments 👍 

re-approving from https://github.com/opendatahub-io/model-registry/pull/3#pullrequestreview-1634759873

antisaling:
> Fixed commit messages and added descriptions. It's an old habit from short single line commit messages from old Apache projects. I'm all for standards and process, but let's make sure it doesn't consume too much of our time and become a chore in itself. 👍

I totally agree with that, I'd just add a small nit now that we use imperative sentences on commit messages (so "added" -> "add"), but if you don't want to fix that in that series it's fine by me.

To be clear, I find that standard makes a lot of sense when you look at git's grammar (and by being consistent we can promote a nicer log)

lampajr (approved):
lgtm

antisaling (approved):
LGTM

dhirajsb:
> I'd just add a small nit now that we use imperative sentences on commit messages (so "added" -> "add"), but if you don't want to fix that in that series it's fine by me.

At this point I've spent more than twice the time fixing commit messages than the time it took to do the actual code changes. Let's merge this PR. :laughing:

PR #4, opened 2023-09-20
Fix golangci-lint local install instructions

lampajr:
## Description
As the `golangci-lint` [documentation](https://golangci-lint.run/usage/install/) suggests I would use `curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.54.2` instead of `go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest`.

Their documentation explicitly says: "Note: such go install/go get installation aren't guaranteed to work. We recommend using binary installation."

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #6, opened 2023-09-20
Reproducible builds

tarilabs:
1. use makefile to automate deps install
2. the .gitignore did not account for file relocations from 97d4c5f

	deleted:    pkg/ml_metadata/proto/metadata_store.pb.go
	deleted:    pkg/ml_metadata/proto/metadata_store_service.pb.go
	deleted:    pkg/ml_metadata/proto/metadata_store_service_grpc.pb.go

bump Go version:

```
 # github.com/sashamelentyev/usestdlibvars/pkg/analyzer/internal/mapping
Error: ../../../go/pkg/mod/github.com/sashamelentyev/usestdlibvars@v1.24.0/pkg/analyzer/internal/mapping/mapping.go:164:7: undefined: time.DateTime
Error: ../../../go/pkg/mod/github.com/sashamelentyev/usestdlibvars@v1.24.0/pkg/analyzer/internal/mapping/mapping.go:165:7: undefined: time.DateOnly
Error: ../../../go/pkg/mod/github.com/sashamelentyev/usestdlibvars@v1.24.0/pkg/analyzer/internal/mapping/mapping.go:166:7: undefined: time.TimeOnly
note: module requires Go 1.20
```

GHA for protoc installation per requirements

## Description
1. setup a GHA action for dry run and reproducible builds according to requirements listed in README (or to be amended)
2. remove files which were supposed not to be checked-in according to `.gitignore`
3. accounts also for: https://github.com/opendatahub-io/model-registry/pull/4#issuecomment-1727157448

## How Has This Been Tested?
see https://github.com/tarilabs/model-registry/actions/runs/6247075284/job/16958971408

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] The developer has manually tested the changes and verified that the changes work: see https://github.com/tarilabs/model-registry/actions/runs/6247075284/job/16958971408

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,31 @@
+name: Build
+
+on:
+  push:
+    branches: [ 'main' ]
+  pull_request:
+    branches: [ '*' ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.19'
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v2
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - name: Install makefile dependencies
+        run: make deps
+      - name: Clean
+        run: make clean
+      - name: Build
+        run: make build
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,14 @@ vet:
 clean:
 	rm -Rf ./model-registry internal/ml_metadata/proto/*.go internal/model/graph/*.go
 
+.PHONY: deps
+deps:
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $$(go env GOPATH)/bin v1.54.2
+	go install github.com/99designs/gqlgen@latest
+	go install github.com/searKing/golang/tools/go-enum@latest
+	go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
+	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
+
 .PHONY: vendor
 vendor:
 	go mod vendor


.gitignore
```diff
@@ -1,6 +1,6 @@
 .idea
 .model-registry.yaml
 model-registry
-ml_metadata/proto/*.go
+pkg/ml_metadata/proto/*.go
```
lampajr: This is probably going to conflict/be wrong wrt https://github.com/opendatahub-io/model-registry/pull/3 so based on which one is going to be merged first, the other one should be adapted accordingly

README.md
```diff
@@ -2,7 +2,7 @@
 A go based server that implements a gRPC interface for [ml_metadata](https://github.com/google/ml-metadata/) library.
 It adds other features on top of the functionality offered by the gRPC interface.
 ## Pre-requisites:
-- go >= 1.19
+- go >= 1.20
```
lampajr: If we change this, shouldn't we change the go version in `go.mod` as well?

Apart from this, is it fine to have precondition on `go >= 1.20` or for some reason should we keep `1.19`?

lampajr (commented):
Thanks a lot @tarilabs for the idea of automating the build and making it reproducible!

Added just some comments/doubt below

.gitignore
tarilabs: Yes, depending on which get merged first, the other will need rebase and mod

README.md
tarilabs: I could not reproduce with Go 1.19, so I would say this must bump or else the code which is already checked in is actually broken

lampajr: yeah I think the issue was during this installation: `go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest` (or maybe not just this one) so if for some reason we need to keep `1.19` we might think to set older dependency version instead of `latest`.

Another question is: is it safe to use the `latest` version for those deps? I do not know how fast Go world is going so we might be in the same situation sooner if they keep releasing new versions (not backward compatible) frequently. Maybe this is not the right moment for this discussion :)

tarilabs: For the "latest" I believe we can keep as-is for the time being, and to get started with. We can always improve on that later (and separate PR)

Makefile
```diff
@@ -21,6 +21,13 @@ vet:
 clean:
 	rm -Rf ./model-registry pkg/ml_metadata/proto/*.go pkg/model/graph/*.go
 
+.PHONY: deps
+deps:
+	go install github.com/99designs/gqlgen@latest
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
+	go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
+	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
```
antisaling: We should probably add the `go-enum` dependency here as well as per [Andrea's note on #3](https://github.com/opendatahub-io/model-registry/pull/3#discussion_r1331245948).

Makefile
```diff
@@ -21,6 +21,13 @@ vet:
 clean:
 	rm -Rf ./model-registry pkg/ml_metadata/proto/*.go pkg/model/graph/*.go
 
+.PHONY: deps
+deps:
+	go install github.com/99designs/gqlgen@latest
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
```
antisaling: Also, this conflicts with the new instructions from #4.

README.md
antisaling: If the code doesn't really work with 1.19, I'd prefer to have a version bump change in another commit so we can refer to it as a separate bugfix.

tarilabs: > If the code doesn't really work with 1.19, I'd prefer to have a version bump change in another commit so we can refer to it as a separate bugfix.

it was, but code-policy requires squashed commit, and reproducible build is the subject here (including, in this case, the version bump)

antisaling (changes requested):
Can you separate the version bump (as commit [1/4]), from the gitignore changes (as commit [2/4]), add the new Makefile rules (as commit [3/4]), and then finally add the CI? This way we can keep nice atomic changes.

Also, as mentioned in [my comment on #3](https://github.com/opendatahub-io/model-registry/pull/3#pullrequestreview-1635508625) we should stick to git's commit conventions.

For some reason I couldn't find a CI run [on your branch](https://github.com/tarilabs/model-registry/tree/tarilabs-20230920-dryrunWithCI), can you provide a link to a successful run with the suggested `build.yml`?

Either way, nice work here, let's keep project maintenance in check :)

Makefile
tarilabs: I agree, but depending on which PR gets in first, we will rebase/modify accordingly.

Makefile
tarilabs: I agree, but depending on which PR gets in first, we will rebase/modify accordingly.

tarilabs:
> Can you separate the version bump (as commit [1/4]), from the gitignore changes (as commit [2/4]), add the new Makefile rules (as commit [3/4], and then finally add the CI? This way we can keep nice atomic changes.

no i cannot. as explained code policy wants PR with squashed commits 🤷 
it was the case before.

> 
> Also, as mentioned in [my comment on #3](https://github.com/opendatahub-io/model-registry/pull/3#pullrequestreview-1635508625) we should stick to git's commit conventions.

see above, the squashed commit text respects the result of that once squashed

> 
> For some reason I couldn't find a CI run [on your branch](https://github.com/tarilabs/model-registry/tree/tarilabs-20230920-dryrunWithCI), can you provide a link to a successful run with the suggested `build.yml`?

I linked it in the original description:

> How Has This Been Tested?
> see https://github.com/tarilabs/model-registry/actions/runs/6247075284/job/16958971408

antisaling:
> > Can you separate the version bump (as commit [1/4]), from the gitignore changes (as commit [2/4]), add the new Makefile rules (as commit [3/4], and then finally add the CI? This way we can keep nice atomic changes.
> 
> no i cannot. as explained code policy wants PR with squashed commits 🤷 it was the case before.

Hmm, about the code policy, does that make sense? I wouldn't say so, as it's a lot easier to keep atomic changes. Either way, I interpret it differently: "the commits are squashed" as in you aren't making a change then fixing it in an upcoming commit in the same series -- so we don't have to refer to multiple changes for a single atomic modification.

EDIT: A little thought experiment: if we were to squash all commits in every PR, then we'd either get massive changes on most PRs or we'd have to separate everything in multiple PRs, either way, this isn't helpful at all for reviews or for the project history.

> > Also, as mentioned in [my comment on #3](https://github.com/opendatahub-io/model-registry/pull/3#pullrequestreview-1635508625) we should stick to git's commit conventions.
> 
> see above, the squashed commit text respects the result of that once squashed
> 
> > For some reason I couldn't find a CI run [on your branch](https://github.com/tarilabs/model-registry/tree/tarilabs-20230920-dryrunWithCI), can you provide a link to a successful run with the suggested `build.yml`?
> 
> I linked it in the original description:
> 
> > How Has This Been Tested?
> > see https://github.com/tarilabs/model-registry/actions/runs/6247075284/job/16958971408

Sorry, I missed that, my bad.

tarilabs:
> Hmm, about the code policy, does that make sense? I wouldn't say so, as it's a lot easier to keep atomic changes (...)

I am empathic with this, but on other hand I don't make the rules 😬 
I think for the time being it is what it is, and we can ask for derogation in using "squash and merge" on github.

> Sorry, I missed that, my bad

No worries.
I would like to proceed swiftly and improve on process progressively.
I noted the "build is broken", this fixes all the necessary broken points and ensure non-regression with the GHA.
Hope this ☝️  is an alternative good "title" if preferred.

README.md
dhirajsb: Let's go with changes from @lampajr . As far as we can, we should try and support the last 3 go releases, which is 1.19, since 1.21 is the latest right now.

tarilabs:
rebased to account for #4

.gitignore
dhirajsb: I removed this ignore. We should check whether any generated code changed before merging PRs.

.github/workflows/build.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - name: Install dependencies
+        run: make deps
+      - name: Clean
+        run: make clean
+      - name: Build
+        run: make build
```
dhirajsb, line 31: Let's add [uncommitted code check](https://github.com/marketplace/actions/check-uncommitted-changes) in there as well.

dhirajsb (dismissed):
lgtm, we can add more checks like [uncommitted code check](https://github.com/marketplace/actions/check-uncommitted-changes) in another PR.

.github/workflows/build.yml
```diff
@@ -0,0 +1,31 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.20.x'
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v2
```
lampajr, line 19: In order to avoid hitting rate-limit we should provide repo-token as suggested https://github.com/arduino/setup-protoc/issues/6#issuecomment-617149930
```suggestion
        uses: arduino/setup-protoc@v2
        with:
          repo-token: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
```
lampajr: Based on their [doc](https://github.com/golangci/golangci-lint-action#how-to-use) it seems that `version` is mandatory. Furthermore it seems that the action is expecting `.golangci.yml` at the root of the repository but it is not clear to me if that MUST exist or not.

```suggestion
        uses: golangci/golangci-lint-action@v3
        with:
          version: v1.54
```

lampajr (changes requested):
Some suggestion, moreover I tried to run the workflow locally but it keeps failing for this:

```
:error::Failed to run: Error: failed to get action config: connect ENETUNREACH 2606:50c0:8002::154:443, Error: failed to get action config: connect ENETUNREACH 2606:50c0:8002::154:443%0A    at getConfig (/run/act/actions/golangci-golangci-lint-action@v3/dist/run/index.js:67135:15)%0A    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)%0A    at async findLintVersion (/run/act/actions/golangci-golangci-lint-action@v3/dist/run/index.js:67156:20)%0A    at async prepareLint (/run/act/actions/golangci-golangci-lint-action@v3/dist/run/index.js:66704:27)%0A    at async prepareEnv (/run/act/actions/golangci-golangci-lint-action@v3/dist/run/index.js:66765:22)
```

It seems something related to the golangci-lint action but I did not find yet the root cause.

.github/workflows/build.yml
tarilabs: well spotted thank you

Makefile
```diff
@@ -21,6 +21,13 @@ vet:
 clean:
 	rm -Rf ./model-registry internal/ml_metadata/proto/*.go internal/model/graph/*.go
 
+.PHONY: deps
+deps:
+	go install github.com/99designs/gqlgen@latest
+	go install github.com/searKing/golang/tools/go-enum@latest
+	go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
+	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
```
lampajr, line 30:
```suggestion
	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b `go env GOPATH`/bin v1.54.2
```

README.md
lampajr:
```suggestion
- go >= 1.19
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,35 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.20.x'
```
lampajr:
```suggestion
          go-version: '1.19'
```

lampajr (changes requested):

Makefile
tarilabs: solved using `$$`

.github/workflows/build.yml
tarilabs: done with bbd018a

tarilabs:
To highlight why I was suggesting not to check-in to version-control the generated files, the check fails because anyone might have a slightly different environment:

```diff
diff --git a/internal/ml_metadata/proto/metadata_store.pb.go b/internal/ml_metadata/proto/metadata_store.pb.go
index 0d89cd2..ee2d0db 100644
--- a/internal/ml_metadata/proto/metadata_store.pb.go
+++ b/internal/ml_metadata/proto/metadata_store.pb.go
@@ -16,7 +16,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 //     protoc-gen-go v1.31.0
-//     protoc        v3.20.3
+//     protoc        v4.24.2
 // source: ml_metadata/proto/metadata_store.proto
 
 package proto
diff --git a/internal/ml_metadata/proto/metadata_store_service.pb.go b/internal/ml_metadata/proto/metadata_store_service.pb.go
index fadd1ce..64ef5ef 100644
--- a/internal/ml_metadata/proto/metadata_store_service.pb.go
+++ b/internal/ml_metadata/proto/metadata_store_service.pb.go
@@ -16,7 +16,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 //     protoc-gen-go v1.31.0
-//     protoc        v3.20.3
+//     protoc        v4.24.2
 // source: ml_metadata/proto/metadata_store_service.proto
 
 package proto
diff --git a/internal/ml_metadata/proto/metadata_store_service_grpc.pb.go b/internal/ml_metadata/proto/metadata_store_service_grpc.pb.go
index cd3945f..7661927 100644
--- a/internal/ml_metadata/proto/metadata_store_service_grpc.pb.go
+++ b/internal/ml_metadata/proto/metadata_store_service_grpc.pb.go
@@ -16,7 +16,7 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
 // - protoc-gen-go-grpc v1.3.0
-// - protoc             v3.20.3
+// - protoc             v4.24.2
 // source: ml_metadata/proto/metadata_store_service.proto
 
 package proto
```

.github/workflows/build.yml
antisaling: I don't think I understand the purpose of this action, can someone explain why we want it?

.gitignore
```diff
@@ -3,3 +3,4 @@
 model-registry
 metadata.sqlite.db
 vendor
+.DS_Store
```
antisaling: While I appreciate this as a Mac user, you should probably ignore that in your ~/.gitignore file (or the project's local .git/info/exclude)

antisaling:
> To highlight why I was suggesting not to check-in to version-control the generated files, the check fails because anyone might have a slightly different environment:

I definitely agree with you. We should only make it easy to regen those. I still haven't tried my hand at that though, do we have any reasons to avoid it?

.github/workflows/build.yml
tarilabs: @isinyaaa it's in relation to the current choice of checking-in generated files, see https://github.com/opendatahub-io/model-registry/pull/6#issuecomment-1729425827

antisaling: IIUC this will inform us about how the CI changed those files then?

tarilabs: the current status of this PR, the GHA checks will inform if the generate (on CI) results in uncommitted changes (I didn't use the linked GHA but just Bash, as the GHA action did not work and is likely unmaintained)

dhirajsb:
We want to make sure the binaries generated are exactly the same using exactly the same tools. 

This is especially important when eventually this becomes a managed service. 
SREs hate having to setup a development environment/CI doing it like in the git actions, and they don't like environment specific side effects during builds. 

So we should have a `bin` directory in the project where we install tools in the Makefile. We can use targets like:
```
bin/golangci-lint:
    curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ./bin v1.54.2
```
And make these tools dependencies of other targets like gen/* for example. 
For the go install commands, we can set the GOBIN directory like so:
```
GOBIN=`pwd`/bin go install github.com/searKing/golang/tools/go-enum@v1.2.97
```
And lock down the tool versions to ensure repeatability. 

@tarilabs if you wish you can do it in this PR, or I can create another PR if you are tired of working on this. :smile:

dhirajsb:
@lampajr @isinyaaa do you have blocking concerns about this PR? If they are things we can improve or sort out later, please remove the change request so we can get these changes in soon.

antisaling:
I'll approve once it's properly rebased and commits are looking good. Even if we're squashing everything the description wouldn't make any sense as is rn.

.gitignore
tarilabs: Most the projects I've worked on put this in the project's `.gitignore` so that doesn't "expect" contributors to have additional local-setup (say someone raise a PR, is likely they might not notice this slips-in).

Example:
- https://github.com/quarkusio/quarkus/blob/main/.gitignore#L1
- https://github.com/apache/camel/blob/main/.gitignore#L6

With the idea is that you make the life easier for any contributors.

But seems to me this is what is preferred to the group, acting on it 👍

tarilabs:
PR reworked per f2d9469 description and externalized "uncommitted changes check" to be dealt later in https://github.com/opendatahub-io/model-registry/issues/16

lampajr (approved):

antisaling (approved):

tarilabs:
on no-veto(s) remaining, and neutral or approved reviews, proceeding to merge.

Issue #14, opened 2023-09-21
[feature] Create a test suite runner for Python test clients


dhirajsb:
**Is your feature request related to a problem? Please describe.**
Need a test runner for a Python test suite to run one or more Python test client scripts against the server. 

**Describe the solution you'd like**
A simple way to run integration tests using something like `make test/integration` is needed. Python probably has a JUnit style test runner facility that can run one or more Python tests. It will be useful if the runner can also start a test go server and shutdown at the end of integration tests. 

**Describe alternatives you've considered**
Current alternative of running a single Python test client is not very useful. 

**Additional context**
Integration tests are more useful at this point than unit tests to make sure all the major happy paths and scenarios for calling the server APIs work as intended.

dhirajsb:
@isinyaaa perhaps you are best suited to start this with your previous Python experience.

dhirajsb:
You probably already knew about it but if you didn't check out [behave](https://behave.readthedocs.io/en/latest/) for BDD in Python.

rareddy:
@isinyaaa have you figured out how you want to write the behavior tests and what also what @dhirajsb asking above?

tarilabs:
Sorry to interject, I'd suggest to possibly consider Robot Framework for BDD as well for the Python SDK (as in addition to User Story over REST), as the initial findings in https://github.com/opendatahub-io/model-registry/pull/108#pullrequestreview-1701740430 are promising

rareddy:
Is this can be closed  @isinyaaa ??

PR #18, opened 2023-09-22
Add basic Unit Test for PutArtifactType logic

tarilabs:
introduce basic go test Unit Test
for testing logic of gRPC PutArtifactType

## How Has This Been Tested?
```
$ go test -timeout 30s -run PutArtifactType github.com/opendatahub-io/model-registry/internal/server -v
=== RUN   TestPutArtifactType
    grpc_server_test.go:63: Should PutArtifactType: type_id:1
    grpc_server_test.go:66: Should PutArtifactTypeResponse.TypeId an ID > 0: type_id:1
--- PASS: TestPutArtifactType (0.01s)
=== RUN   TestPutArtifactTypeNoNameFailure
    grpc_server_test.go:95: Should error on PutArtifactType with Request missing Name: rpc error: code = Internal desc = server panic: runtime error: invalid memory address or nil pointer dereference
--- PASS: TestPutArtifactTypeNoNameFailure (0.00s)
PASS
ok  	github.com/opendatahub-io/model-registry/internal/server	0.608s
```

## Merge criteria:

- [x] The developer has manually tested the changes and verified that the changes work (see above)

### DIFF
--- a/internal/server/grpc_server_test.go
+++ b/internal/server/grpc_server_test.go
@@ -62,10 +62,9 @@ func TestPutArtifactType(t *testing.T) {
 	}
 	t.Logf("Should PutArtifactType: %v", response)
 
-	if *response.TypeId > 0 {
-		t.Logf("Should PutArtifactTypeResponse.TypeId an ID > 0: %v", response)
-	} else {
-		t.Errorf("Should PutArtifactTypeResponse.TypeId an ID > 0: %v", response)
+	t.Logf("Should PutArtifactTypeResponse.TypeId an ID > 0: %v", response)
+	if *response.TypeId <= 0 {
+		t.Fail()
 	}
 }
 
--- a/internal/server/grpc_server_test.go
+++ b/internal/server/grpc_server_test.go
@@ -93,6 +93,6 @@ func TestPutArtifactTypeNoNameFailure(t *testing.T) {
 	if err != nil {
 		t.Logf("Should error on PutArtifactType with Request missing Name: %v", err)
 	} else {
-		t.Errorf("Should error on PutArtifactType with Request missing Name: %v, %v", response, err)
+		t.Errorf("Should error on PutArtifactType with Request missing Name: %v", response)
 	}
 }
--- a/internal/server/grpc_server_test.go
+++ b/internal/server/grpc_server_test.go
@@ -1,0 +1,99 @@
+package server
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"os"
+	"testing"
+
+	"github.com/opendatahub-io/model-registry/internal/ml_metadata/proto"
+	"github.com/opendatahub-io/model-registry/internal/model/db"
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+)
+
+func migrateDatabase(dbConn *gorm.DB) error {
+	// using only needed RDBMS type for the scope under test
+	err := dbConn.AutoMigrate(
+		db.Type{},
+	)
+	if err != nil {
+		return fmt.Errorf("db migration failed: %w", err)
+	}
+	return nil
+}
+
+func setup(tmpFile *os.File) (*gorm.DB, error) {
+	db, err := gorm.Open(sqlite.Open(tmpFile.Name()), &gorm.Config{})
+	if err != nil {
+		return nil, err
+	}
+	err = migrateDatabase(db)
+	if err != nil {
+		return nil, err
+	}
+	return db, nil
+}
+
+// Bare minimal test of PutArtifactType with a given Name
+func TestPutArtifactType(t *testing.T) {
+	f, err := os.CreateTemp("", "model-registry-db")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer os.Remove(f.Name())
+	db, err := setup(f)
+	if err != nil {
+		t.Errorf("Should expect DB connection: %v", err)
+	}
+	grpcServer := NewGrpcServer(db)
+	artifactTypeName := "test"
+	artifactType := proto.ArtifactType{
+		Name: &artifactTypeName,
+	}
+	req := proto.PutArtifactTypeRequest{
+		ArtifactType: &artifactType,
+	}
+
+	response, err := grpcServer.PutArtifactType(context.Background(), &req)
+	if err != nil {
+		t.Errorf("Should PutArtifactType: %v", err)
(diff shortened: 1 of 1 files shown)


internal/server/grpc_server_test.go
```diff
@@ -0,0 +1,99 @@
...
+		Name:    nil,
+		Version: &versionName,
+	}
+	req := proto.PutArtifactTypeRequest{
+		ArtifactType: &artifactType,
+	}
+
+	response, err := grpcServer.PutArtifactType(context.Background(), &req)
+	if err != nil {
+		t.Logf("Should error on PutArtifactType with Request missing Name: %v", err)
```
lampajr, line 95:
```suggestion
		t.Logf("Expected error on PutArtifactType with Request missing Name: %v", err)
```

lampajr (dismissed):
lgtm

internal/server/grpc_server_test.go
```diff
@@ -0,0 +1,99 @@
...
+// Bare minimal test of PutArtifactType with a given Name
+func TestPutArtifactType(t *testing.T) {
+	f, err := os.CreateTemp("", "model-registry-db")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer os.Remove(f.Name())
+	db, err := setup(f)
+	if err != nil {
+		t.Errorf("Should expect DB connection: %v", err)
```
isinyaaa, line 48: `Should expect` in error messages sounds confusing. How about:

```suggestion
		t.Errorf("Expected DB connection: %v", err)
```

internal/server/grpc_server_test.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		t.Errorf("Should PutArtifactType: %v", err)
+	}
+	t.Logf("Should PutArtifactType: %v", response)
+
+	if *response.TypeId > 0 {
+		t.Logf("Should PutArtifactTypeResponse.TypeId an ID > 0: %v", response)
+	} else {
+		t.Errorf("Should PutArtifactTypeResponse.TypeId an ID > 0: %v", response)
+	}
```
antisaling: If both messages are the same, why don't we
```suggestion
	t.Logf("Should PutArtifactTypeResponse.TypeId an ID > 0: %v", response)
	if *response.TypeId <= 0 {
		t.Fail()
	}
```

Reference: https://pkg.go.dev/testing#B.Errorf

internal/server/grpc_server_test.go
```diff
@@ -0,0 +1,99 @@
...
+	req := proto.PutArtifactTypeRequest{
+		ArtifactType: &artifactType,
+	}
+
+	response, err := grpcServer.PutArtifactType(context.Background(), &req)
+	if err != nil {
+		t.Logf("Should error on PutArtifactType with Request missing Name: %v", err)
+	} else {
+		t.Errorf("Should error on PutArtifactType with Request missing Name: %v, %v", response, err)
```
antisaling: Here, `err` should be `nil`, so why print it?

antisaling (commented):
lgtm aside from error messages. I don't get why we're using "should ..." here every time. Some of them don't make any sense to me.

Also: please use the imperative on the commit message.

internal/server/grpc_server_test.go
tarilabs: I just adopted "test styleguide" from a course I've followed which tends to reuse the verbatim but use Logf Vs Errorf 🤷

tarilabs:
> I don't get why we're using "should ..." here every time. Some of them don't make any sense to me.

👉  https://github.com/opendatahub-io/model-registry/pull/18#discussion_r1334532493

> Also: please use the imperative on the commit message.

I have used title within 50 chars, newline, description; I am afraid I did not understand what you meant here. Do you have an article to point me to, please.

internal/server/grpc_server_test.go
antisaling: I tend to be against this because DRY on string makes a lot of sense -- one less place to update in case the test changes, and it's not totally obvious that both strings are the exact same.

antisaling:
> > I don't get why we're using "should ..." here every time. Some of them don't make any sense to me.
> 
> 👉 [#18 (comment)](https://github.com/opendatahub-io/model-registry/pull/18#discussion_r1334532493)

EDIT: I see it now. Can you link to this style guide so we can all understand where you're coming from? @tarilabs 

> > Also: please use the imperative on the commit message.
> 
> I have used title within 50 chars, newline, description; I am afraid I did not understand what you meant here. Do you have an article to point me to, please.

This one is pretty good: https://cbea.ms/git-commit/
@lampajr @dhirajsb if you haven't read it, please check it out as well

tarilabs:
> Also: please use the imperative on the commit message.

👉  21743b5

dhirajsb:
In general unit tests for each API are not a good idea. They have a very limited scope and therefore a very short period where they are useful. They can also cause a lot of repetition in test setup. 

For the OpenShift Connectors Control Plane API, we built an extensive suite of BDD style integration API call  _scenarios_, which tested multiple APIs in various sequences. Those tests were a lot more useful because they exercised APIs in real world use cases and served as excellent regression tests. 
We don't have to create an extensive test suite and all the scenarios right away, but we should start building a test runner like I mentioned in #14. We can start with simple scenarios that look like unit tests, but grow and become more complex over time.

internal/server/grpc_server_test.go
tarilabs: I will adopt this other "styleguide" as I also tend to agree with DRY, until I get more clear the majority preference for the group

lampajr (approved):

dhirajsb (changes requested):
Re-use this as a unit test for the DB service handler. That way it's a quick and easy unit test and we leave integration testing to python scripts.

antisaling:
The last commit could've been a fixup. If you do squash this, please remove that message.

tarilabs:
> Re-use this as a unit test for the DB service handler. That way it's a quick and easy unit test and we leave integration testing to python scripts.

I think we discussed about having small unit test for "surface contacts" or "core capabilities" and serving as smoke testing was helpful.
This could have served as a small smoke/unit test in that sense.
It seems however that there is no way to have a change-of-mind about this.
Noted. ✍️ 

Rejected with https://github.com/opendatahub-io/model-registry/pull/18#pullrequestreview-1644747578 , closing PR.

PR #34, opened 2023-09-26
Add first draft of Data Access Layer, with test

tarilabs:
## Description
Pair programmed with @lampajr as discussed first draft of Data Access Layer (DAL).

Key points to be discussed:

- we opted for multiple `<entity>Handler`(s) interface to leverage GoLang implicit interface on call site if ever needed
    - we considered "one giant interface" but we preferred above
- we opted during `create` method of CRUD, to explicit all properties of the entity, as from client call site only after creation you get an instance of the entity, which could be passed to Update, Delete, etc.
    - we considered passing entity also for creation method but we preferred above
    - there is some opportunity for further isolation of DAL with regards to "built-in types" (int, bool, string, struct, proto, etc) as defined by the proto, for the time being we opted to keep this in order to make it explicit (as anyway a change in the proto will need to reflect on the DAL anyway)
    - we provide convenience query method for what we believe are most common needed

Open points:
- where to place logic for attributes, for example unique version for an artifactType: this DAL service or directly inside DB as constraint/key ?
- semantic of `Create` in CRUD: keep currently existing Create-or-Update-if-already-exising or switch to Create-or-Error-if-already-existing ?

## How Has This Been Tested?
```
go test -timeout 30s ./...
```

## Merge criteria:


lampajr (dismissed):

internal/service/db_service.go
```diff
@@ -0,0 +1,150 @@
...
+
+type ArtifactTypeHandler interface {
+	CreateArtifactType(name string, version *string) (result *db.Type, err error)
+	CreateArtifactTypeAll(name string,
+		version *string,
+		description *string,
+		inputType *string,
+		outputType *string,
+		externalID *string,
+		properties map[string]proto.PropertyType) (result *db.Type, err error)
```
rareddy: - This call is always behind the gRPC call, IMO we should always use `db.ArtifactType` as the input nd not have individual properties defined. The converters for each type are being generated where gRPC -> DB.
- Since we are defining the database layer, how about using methods corresponding to its lingo like `insert`, `read`, `update`, and `delete`.  Can also have an `upsert` which is `insert` when no record found otherwise a `update`?
- Since we are saying individual interface for each table, can we drop the verbose `Artifact` in the method names?

internal/service/db_service.go
```diff
@@ -0,0 +1,150 @@
...
+	GetArtifactTypes(id *int64, name *string, version *string) ([]db.Type, error)
+	GetArtifactType(id *int64, name *string, version *string) (*db.Type, error)
+
+	DeleteArtifactTypesByName(name string) ([]db.Type, error)
+	DeleteArtifactTypeByNameVersion(name string, version string) (*db.Type, error)
+	UpdateArtifactType(artifactType db.Type) (*db.Type, error)
+}
+
+type ArtifactHandler interface {
```
rareddy: One interface per `go` file?

internal/service/db_service.go
```diff
@@ -0,0 +1,150 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"gorm.io/gorm"
+
+	"github.com/opendatahub-io/model-registry/internal/ml_metadata/proto"
+	"github.com/opendatahub-io/model-registry/internal/model/db"
+	"github.com/opendatahub-io/model-registry/internal/server"
+)
+
+type ArtifactTypeHandler interface {
```
dhirajsb: Create a single interface called DBService with Create/Update functions for all schema tables.

internal/service/db_service.go
```diff
@@ -0,0 +1,150 @@
...
+type TypeKind int32
+
+// TODO: move from grpc to just here
+const (
+	EXECUTION_TYPE TypeKind = iota
+	ARTIFACT_TYPE
+	CONTEXT_TYPE
+)
+
+type Handle struct {
```
dhirajsb: Move the implementation into another file `db_service_handler.go`.

internal/service/db_service.go
```diff
@@ -0,0 +1,150 @@
...
+	"google.golang.org/grpc/status"
+	"gorm.io/gorm"
+
+	"github.com/opendatahub-io/model-registry/internal/ml_metadata/proto"
+	"github.com/opendatahub-io/model-registry/internal/model/db"
+	"github.com/opendatahub-io/model-registry/internal/server"
+)
+
+type ArtifactTypeHandler interface {
+	CreateArtifactType(name string, version *string) (result *db.Type, err error)
```
dhirajsb: All DBService funcs should take db.* structs as parameters. Since that's the output of the gRPC to DB Converters. 
That will allow us to work with not only all the fields in that type, but since we have relationships in them, the `DBServiceHandler` can also use Gorm [associations](https://gorm.io/docs/associations.html) to automatically create/update `Properties` (MUST be done) and other relationships like `attributions` and `associations`. 
There are some APIs in the gRPC interface where multiple objects are created in a single call (A `Context` with all it's `Artifacts` and `Executions`). So the DBService needs to be able to support that as well. Gorm associations will make that easy to do. Just need to check for the presence of the relationship field, and make sure foreign keys are correctly set before saving the association. There is also a `FullSaveAssociation` option which can save an entire tree of objects.

internal/service/db_service.go
dhirajsb: Create a single interface for all types called `DBService`, otherwise we will have a clutter of DB interfaces and handlers to deal with. 
The implementation can still be split into multiple handlers using anonymous fields like so:
```go
// make sure handler implements all interface operations
var _ DBService = dbServiceHandler{}
var _ DBService = (*dbServiceHandler)(nil)

func NewDBService() DBService {
  return &dbServiceHandler{}
}

struct dbServiceHandler {
  artifactHandler
  contextHandler
  ...
}
```
Then split out `*Handler structs` into `*_handler.go`. This way the service user just needs to know about `DBService` and `NewDBService`, which are the only types exported. 
If the individual handlers need to call other handlers (to handle associations), then they could have a `DBService` field, which could be initialized to `*dbServiceHandler` in `NewDBService`.

dhirajsb (changes requested):
Added comments for required changes.

internal/service/db_service.go
```diff
@@ -0,0 +1,150 @@
...
+	CreateArtifactTypeAll(name string,
+		version *string,
+		description *string,
+		inputType *string,
+		outputType *string,
+		externalID *string,
+		properties map[string]proto.PropertyType) (result *db.Type, err error)
+	GetArtifactTypes(id *int64, name *string, version *string) ([]db.Type, error)
```
dhirajsb: Get functions should use a signature similar to the gorm [Where](https://github.com/go-gorm/gorm/blob/e57e5d8884d801caa4ce0307bcd081f7e889e514/chainable_api.go#L200) func like so:
```go
GetXYZ(query interface{}, args ...interface{}) ([]db.XYZ, err)
```
This will make the interface really flexible, and allows users of the interface to choose query and parameters. Granted it's a very thin layer, but it works for now. We can figure out what our query requirements are going to be as we tackle more complex use cases in the future and refactor this function as needed.

antisaling:
@lampajr are you sure you want you personal email in that `co-authored-by` tag? no problem by me, just a heads up

lampajr:
> @lampajr are you sure you want you personal email in that `co-authored-by` tag? no problem by me, just a heads up

@isinyaaa Thanks for the heads up but that is fine by me too, as this is the personal GitHub account I always used the primary email (which is the personal one)

internal/service/db_service_test.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		t.Errorf("Should create ArtifactType: %v", err)
+	}
+	at1, err := dal.CreateArtifactType(artifactName, &v1)
+	if err != nil {
+		t.Errorf("Should create ArtifactType: %v", err)
+	}
+	if at0.ID > at1.ID {
+		t.Errorf("ID invariant does not hold")
+	}
```
tarilabs: @lampajr just for the record this I realized was not behaving as intended, as `at0` and `at1` did in fact had the same ID (the invariant we wanted to check I think was having diff id, and for v1 to be > than v0, but it's not actually the case because https://github.com/opendatahub-io/model-registry/pull/36#pullrequestreview-1648369631)

internal/server/db_context.go
```diff
@@ -1,8 +1,10 @@
+// TODO when refactoring to migrate to service/db_service usage, this should move inside service/ too.
```
dhirajsb: Right, it should be moved to `internal/service/db_context.go`. Feel free to do so now.

internal/service/db_service.go
```diff
@@ -0,0 +1,48 @@
...
+
+	// InsertEEE(db.EEE) (*db.EEE, error)
+	// UpsertEEE(db.EEE) (*db.EEE, error)
+	// ReadEEE(db.EEE) (*db.EEE, error)
+	// ReadAllEEE(query interface{}, args ...interface{}) ([]*db.EEE, error)
+	// UpdateEEE(db.EEE) (*db.EEE, error)
+	// DeleteEEE(db.EEE) (*db.EEE, error)
+}
+
+type TypeKind int32
```
dhirajsb: We can reuse the enum from `db/type.go`.

internal/service/db_service_test.go
```diff
@@ -0,0 +1,142 @@
...
+		db.TypeProperty{},
+		// TODO: add as needed.
+	)
+	if err != nil {
+		return fmt.Errorf("db migration failed: %w", err)
+	}
+	return nil
+}
+
+func setup(tmpFile *os.File) (*gorm.DB, error) {
```
dhirajsb: No need for a tmpFile, SQLite supports an [in-memory DB](https://www.sqlite.org/inmemorydb.html).

internal/service/db_service_tx.go
```diff
@@ -0,0 +1,27 @@
+package service
```
dhirajsb: This can be moved into `internal/service/db_context.go`.

internal/server/db_context.go
tarilabs: Not really "free", I'm still prevented from merging this PR.

internal/service/db_service_tx.go
tarilabs: c4e72ff

internal/server/db_context.go
tarilabs: c4e72ff

internal/service/db_service.go
tarilabs: 0b0d5ca

internal/service/db_service_test.go
tarilabs: 👉  I do not share the same opinion as this requested change will prevent from running test in parallel (same DB "in memory" is used) but anyway 🤷 

/cc @lampajr to be noted ✍️ 

062e9a8

dhirajsb: SQLite doesn't share the in-memory DB, every call (even with the same connection string) creates a new in-memory DB.

dhirajsb (approved):
lgtm, just one small nitpick. Use non-sql operation verbs for services like these. So, `Create*` instead of `Insert*`, and `Get*` instead of `Read*`.

internal/service/db_service_test.go
tarilabs: > SQLite doesn't share the in-memory DB, every call (even with the same connection string) creates a new in-memory DB.

Again, that is not my understanding.

GORM manual suggest you use the connection string for memory sharing since GORM have connection pooling; so this is the reason of the function deferred to close up all connections, hence clearing the DB for the next one.

I've also tried using just `:memory:` so going _different_ from what advised in the GORM manual, and it has the downside one would expect: since as a downstream user you have no control over the connection, during test at times the DB is "cleared" from one orm call to the next one (specifically, DB tables were absent in subsequent calls).

tarilabs:
> (...) just one small nitpick. Use non-sql operation verbs for services like these. So, Create* instead of Insert*, and Get* instead of Read*.

We actually took care to implement as originally mentioned here: https://github.com/opendatahub-io/model-registry/pull/34#discussion_r1337335268

lampajr (approved):

tarilabs:
No longer vetos, approvals, proceeding to merge.

Issue #33, opened 2023-09-25
[feature] Externalize the DDL for the Database schema


rareddy:
**Is your feature request related to a problem? Please describe.**

Right now the code to generate the database schema is embedded inside the `gorm` classes, however, it is preferable to externalize this and use a solution like `liquidbase` for version management. There may be different equivalent libraries for Go

**Describe the solution you'd like**

@dhirajsb mentioned a tool `Atlas`, for the needs of this. We need to do a spike on this feature.

dhirajsb:
Since we have a plan to accomplish this using Gorm migrations support, can we close this issue?

rareddy:
yes, as per discussion on #wg-model-registry the form `migrate` should provide the similar functionality to liquidbase/flyway but the migrations written using `gorm`. The advantage is the code generated will be DB specific and no need to maintain the individual dialects.

@dhirajsb do you want to start at version 10 of the schema as ml-metadata shows, then keep track from there? I wondering how do we make sure all the currently designed indexes and any other quirky DB features are correctly captured in the form metadata when `gorm` DB objects are designed.

dhirajsb:
All the gorm DDL config is captured in `gorm` tags in the DB structs source code. Once we switch to using migrations, we can use the `date` signature like rhosak `migrations.go` and use that to identify schema versions. That way we could have multiple schema migrations captured in a single release cycle. 
It also makes it easier to identify when a particular design change was made, instead of a magical schema version number that has to be externally tied to some release cycle, etc. Does that make sense?

rareddy:
I get that, I am asking how to make sure what gorm has interms of metadata has a complete view of what was showing original projects.

rareddy:
Since will be C++ as is this is no longer required now. Closing as won't do.

PR #36, opened 2023-09-27
Fixed #35, refactor DB schema

### LINKED ISSUE #35: [chore] Cleanup DB schema
The current DB schema was imported from the mlmd project. Some of the column types, relationships, etc. are not optimal. 

* Add missing indexes
* Constraints for keys
* Associations
* Use appropriate column sizes for numeric types
* Set creation and modification timestamps automatically using gorm tags

dhirajsb:
## Description

Gorm changes in internal/model/db:
Add missing indexes
Constraints for keys
Add Associations
Use appropriate column sizes for numeric types
Set creation and modification timestamps automatically using gorm tags

Clean up Makefile to install pinned deps in ./bin

## How Has This Been Tested?
```
make run/server &
make run/client
```

## Merge criteria:

- [X] The developer has manually tested the changes and verified that the changes work


Makefile
```diff
@@ -19,15 +26,15 @@ vet:
 
 .PHONY: clean
 clean:
-	rm -Rf ./model-registry internal/ml_metadata/proto/*.go internal/model/graph/*.go
+	rm -Rf ./model-registry internal/ml_metadata/proto/*.go internal/model/graph/models_gen.go
 
 .PHONY: deps
 deps:
```
lampajr, line 32: Is there a way to manage `protoc` as any other dependency? In that way we will use the same protoc binaries.
Because for instance if I try to run `make clean build` I will have some diffs in the generated grpc stubs, like:
```
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 //     protoc-gen-go v1.31.0
-//     protoc        v3.20.3
+//     protoc        v4.24.2
```

internal/model/db/artifact.go
```diff
@@ -0,0 +1,52 @@
+package db
```
lampajr, line 1: Did you generate all db entities using some automated script? If so, is it something that could be useful in future? 

I am asking this because if it could become useful maybe it would be worth documenting it somewhere.

lampajr (commented):
@dhirajsb overall looks good to me, added just some comments below

bin/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
```
tarilabs, line 1: what about listing `bin` on the root .gitignore and add here a `.gitkeep` file (if we want to persists bin/ inside the repo for stuff which does not mkdirs bin/ if nonexisting)

internal/model/db/artifact.go
tarilabs: I have the same question as the filename change seems to me to suggest we will no longer be generating these entities, also comment that file was autogenerated has been dropped.

Makefile
tarilabs: I have the same issue as previously mentioned 👀 F

antisaling (dismissed):
lgtm, I just want to understand the motivation behind editing the auto-generated files. Is it because those schemas are "fixed" so we won't have to regen them?

Also, please change the commit message, let's keep messages imperative (i.e. `Refactor DB schema` should suffice) then use tags (i.e. `Closes: #35` tag) on the description.

Makefile
dhirajsb: Looks like the version I downloaded from protoc documentation download page is an older version. So, I'll update my version to the latest v4.24.2 that you are using. 
Unfortunately, there is no platform agnostic way to download the binaries for protoc compiler, unless we want to add a huge amount of platform detection code in Makefile. We can pin the version to v4.24.2 in the README, but do you have any other ideas?

internal/model/db/artifact.go
dhirajsb: Yeah, it would have been nice if the schema could be imported as is, but unfortunately the CPP mlmd schema is not directly usable in it's imported form. They don't believe in foreign keys(!!) and some index definitions come in a format that breaks cross DB compatibility. 
The tool I used to import them is [gorm-gen](https://gorm.io/gen/index.html). 
So we could do the same again if we want to fetch an updated mlmd schema.

bin/.gitignore
dhirajsb: Having `.gitignore` keeps our options open, just in case we have to add any helper scripts or something in the `bin` directory. Having a checked in empty directory is one less thing the growing Makefile has to worry about. :smiley:

dhirajsb:
Yes, we don't plan on importing DB schema updates continuously from mlmd, since we added a bunch of foreign keys and indexes that are not in the mlmd schema. 
The gRPC API keeps us forwards compatible. We'll manually update the schema (we could re-import and diff for fields) in the future if we come across a specific use case that requires some change in ml-metadata schema.

internal/model/db/artifact.go
rareddy: I suspect they must have run into some performance issues with Foreign Keys, keeping the database up with ever-increasing data will be a challenge. Most databases do not correctly represent the metadata for indexes correctly especially if they are somewhat exotic.

I will say we do not have any requirement to support unlimited databases. My top picks would be Postgres, SQLite, and RDS, which should cover product versions very decently. This was one of the reasons I was thinking of keeping external DDL might work better but I understand it will not flow with the overall design. Please note there will be a lot of data that gets pumped into these DBs.

bin/.gitignore
tarilabs: That would indeed be the purpose of `.gitkeep` also, don't you agree?
E.g. `bin/.gitkeep`
This would have indeed checkin the bin directory empty as well (as end results) which appears to me you are confirming is indeed the scope here.

tarilabs:
> We'll manually update the schema ...

It seems to me this 👆
somehow is a bit impacting with 👉
- https://github.com/opendatahub-io/model-registry/issues/33

My Q is: I don't fully understand yet if we are looking to
- have a Go struct "first" approach with the DDL generated by Gorm out of it as well as DB migrations,
- or, a "DDL-first" approach (so Go struct will eventually be generated again deriving from that),

please?🤔

internal/model/db/artifact.go
dhirajsb: Yeah, fun fact that you might have noticed already is that there are no delete operations (except for properties) in the gRPC API.

bin/.gitignore
dhirajsb: Git doesn't specifically look for a `.gitkeep` file. It's a hack to add empty directories. But `bin` won't be empty after tools are downloaded. So, now we need to capture that ignore directive somewhere. Which leads to two files, the root `.gitignore` and `.gitkeep` hack. It's a lot easier to just have a  `.gitignore` in `bin`.

dhirajsb:
`Gorm` struct first approach, as far as we can tell right now. The other issue was opened because there was some concern whether `gorm` can handle migrations. That question has been addressed and I've asked to close the other other issue.

Makefile
dhirajsb: I upgraded to the latest version v24.3, and re-generated the files and updated the github action. 
I also added the link to protoc [github release](https://github.com/protocolbuffers/protobuf/releases/tag/v24.3) for that version in README.md. 
Let us all use that version to avoid this issue locally.

tarilabs (approved):
I acknowledge change in strategy from current setup (structs where codegenerated by DB inspection) to new setup "Go struct first" as the source of truth--which will be used to generate via Gorm DDL and DB migrations ref https://github.com/opendatahub-io/model-registry/pull/36#issuecomment-1738099431

internal/model/db/type.go
```diff
@@ -0,0 +1,32 @@
+package db
+
+const TableNameType = "Type"
+
+// Type mapped from table <Type>
+type Type struct {
+	ID          int64   `gorm:"column:id;not null;primaryKey;autoIncrement:true" json:"-"`
+	Name        string  `gorm:"column:name;type:varchar(255);not null;uniqueIndex:idx_type_name,priority:1" json:"-"`
```
tarilabs, line 8: as I'm working on #34 , I realize the `uniqueIndex` is currently prohibiting having 2 (Artifact)Type with same name at different versions. Is this really intentional? 🤔 

Implications:
Artifacts which were stored in relation to a previous version of Type, get automatically "promoted" to the new version of Type once the entry in Type table row is updated.

rareddy: Suggest checking the upstream for how it is defined in DDL.

tarilabs: > Suggest checking the upstream for how it is defined in DDL.

for as much as I could see, iff this is the DDL we're speaking of:

https://github.com/google/ml-metadata/blob/96fb458d22e72f171727f626d7d2f8c779a73ccf/ml_metadata/util/metadata_source_query_config.cc#L41

there is no trace of `UNIQUE(...)` you could see for other tables instead.

But, I might have missed other upstream resources, in this consideration.

rareddy: That is exactly what I was thinking. IMO, we need to keep them in sync, you can extend with additional tables but do not change how they behave otherwise we lose the ability to move previous workloads onto this. For me keeping the external DDL seemed like simplest approach we can take but obviously that brings other issues with `gorm`

dhirajsb: > Implications:
> Artifacts which were stored in relation to a previous version of Type, get automatically "promoted" to the new version of Type once the entry in Type table row is updated.

That is correct. User shouldn't be able to delete properties in a Type, only add new ones (check the gRPC annotations to confirm). That will make types backwards compatible. Not doing so opens a can of worms with Type versioning.

dhirajsb:
I'm going to merge this for now to get the major changes into `main` as there are other PRs and work that depend on this. The schema is not set in stone, we can always keep tweaking the DB schema and in the future write migrations to evolve it.

PR #38, opened 2023-09-27
Fixes #11, add gRPC to/from DB type converter

### LINKED ISSUE #11: [feature][service] Add gRPC to DB type converters
There is a need for converting from types generated from mlmd proto definitions to go structs used by the DB layer. 

Write a go file `converter/proto_converter.go` with functions with the signature:
```
func ConvertProtoArtifactToDB(value *proto.Artifact)  (*db.Artifact, error)
func ConvertDBToProtoArtifact(value *db.Artifact)  (*proto.Artifact, error)
```

The converter should also map any properties that belong to the metadata entity. There is no need to map relationships at this point, e.g. copying `Executions` for `Contexts`. Consumer of the converter will decide whether that's needed and use a converter for related entities. 

This can potentially be done a lot quickly using [goverter](https://github.com/jmattheis/goverter), which takes an interface as an input and generates an implementation based on matching field names. Check to see if it works as it is or whether `goverter` needs some _guidance_ to map certain fields.  

The alternative to using `goverter` would be to hand write all the converter functions. 

**Additional context**
These converters will be part of the DB service layer used for CRUD operations.

dhirajsb:
## Description
Added `internal/converter/grpc_converter.go` that uses goverter to generate a converter in `internal/converter/generated/converter.go`. This converter will be used by `grpc_server.go` to convert to/from proto types and DB.

## How Has This Been Tested?
Includes unit tests for `internal/converter/grpc_converter_util.go` utility functions. 

## Merge criteria:

Another PR for the DB schema refactor needs to be merged before this PR.

- [X] The developer has manually tested the changes and verified that the changes work


dhirajsb:
Ignore the DB changes in this PR since those are from #36 for db schema update.

tarilabs:
> Ignore the DB changes in this PR since those are from #36 for db schema update.

seems 758b8cb to be ignored too as pertaining to another scope?

internal/converter/grpc_converter_util.go
```diff
@@ -0,0 +1,355 @@
...
+
+func ConvertTypeIDToName(id int64) (*string, error) {
+	if id == 0 {
+		return nil, nil
+	}
+	if globalDB == nil {
+		return nil, fmt.Errorf("converter DB connection has not been initialized")
+	}
+	var name string
+	if err := globalDB.Model(db.Type{}).Select("name").Where("id = ?", id).First(&name).Error; err != nil {
```
tarilabs: This functions looks to me is called in conversion (of proto grpc to db) each time based on the annotations in the other file, and contains a DB lookup.

I didn't understand then if this implies the DB lookup is performed at each conversion, and in case, what are the design implications, also because it looks to me is _not_ in a tx-context (DB data race as you might be writing on stale data for type id)?

internal/converter/grpc_converter_util.go
```diff
@@ -0,0 +1,355 @@
...
+		return nil, fmt.Errorf("error getting type name for type id %d: %w", id, err)
+	}
+	return &name, nil
+}
+
+func ConvertToProtoArtifactProperties(source []db.ArtifactProperty) (map[string]*proto.Value, error) {
```
tarilabs, line 54: Shouldn't all these `ConvertToProto<entity>Properties`, `ConvertToProto<entity>CustomProperties`, pairs be made generic in the sum types of `<entity>`s? 🤔 

They seem to me only differing in `p.GetIsCustomProperty()` or the same being negated.

(edit: formatting)

internal/converter/grpc_converter_util_test.go
```diff
@@ -0,0 +1,875 @@
...
+	"gorm.io/gorm"
+	"gorm.io/gorm/logger"
+	"log"
+	"os"
+	"reflect"
+	"testing"
+	"time"
+)
+
+func TestConvertArtifact_State(t *testing.T) {
```
tarilabs, line 17: only this and `TestConvertExecution_State` do not have the `wantErr` column, just mentioning in case you wanted to include for consistency considering in case we will need more rows in future.

tarilabs (commented):
Thank you for providing these unit tests I believe they are also helpful to familiarize on how these converters works.

Noting t.Run() execute straight away so noted in my understanding not impacted by similar issues as, for instance, [here](https://github.com/kubernetes/kubernetes/issues/98213#event-6581848177).

internal/converter/grpc_converter_util.go
dhirajsb: Because we can't cast array types all the functions can't be made generic entirely. I can reduce some code repetition by writing a generic inner body, but the looping on elements of `[]*Properties` still has to be repeated.

internal/converter/grpc_converter_util.go
dhirajsb: Since a user is allowed to create a new type at runtime, this function has to do a DB query for every conversion call. We could potentially optimize this in the future by creating a re-usable cache of id-name map in the DAL(data access layer). 
If a non-zero type ID is provided by a gRPC client, it has to have received it from a committed DB tx in a previous call. So, no race condition there. If the DAL created a new type, the create* DAL call would have committed the TX before returning. So, no race condition again. 
Did I miss any other scenario that you were thinking about?

dhirajsb: And the converter is used from db to proto conversion.

internal/converter/grpc_converter_util.go
dhirajsb: Added a generic `convertToProtoMetadataProperty()`

internal/converter/grpc_converter_util_test.go
dhirajsb: There is no need for `wantErr` for these, since they don't throw errors. 
BTW, `Goland` made it quick and easy to generate the skeleton for these tests. I just had to add the test data.

internal/converter/grpc_converter_util.go
tarilabs: > We could potentially optimize this in the future by creating a re-usable cache of id-name

Agree, Correctness must come first, optimizations later.
Just wanted to double-check I spotted correctly this lookup is on a hot-path, so worth of noting the opportunity for optimization in another issue before merging.

dhirajsb:
Can one of you guys approve this PR, so we can at least merge this code base when we need the go server?

tarilabs:
Didn't you want to record a GH issue to track for the race condition _and_ optimisation aspects, we noticed in https://github.com/opendatahub-io/model-registry/pull/38#discussion_r1339597640 ? 🤔

dhirajsb:
Added a cache to the gRPC converter, since that's the only place it's being used now. We can move it into the DAL later if we need to.

tarilabs:
> 151fd45

isn't this subject to the potential race condition we discussed?
The name conversion function is using a cache with mutex (got it) however the potential [race condition](https://github.com/opendatahub-io/model-registry/pull/38#discussion_r1339597640) is the time which occurs between the data preparation in the converters and the DB write (or read)?

Eg

1. ConvertTypeIDToName thinks ID 1 in DB is "name1", store `1 : name1` in cache
(no tx context for the following steps)
2. other domain objects are converted, using `1 : name1` from cache
3. domain objects get written into DB

between 1 and 3 there is no guarantee the type-id isn't changed on the DB; in my understanding, nothing prevents as you are about to perform step 2 or 3 something else wrote `1: name_1` in DB.
What am I missing? 🤔

dhirajsb:
No, there is no race condition because an id-name association is static after it's created. IDs are auto increment primary keys in the DB, so they won't change. And names are read-only after creation. From ml-metadata proto comments:
```
If a type with the same name exists in the database, it updates the existing type. 
```
So a type name is pretty much synonymous with it's ID.

tarilabs:
> names are read-only after creation

Okay, so my understanding is that since this is an invariant we are going to assume/respect while interfacing with the DB (eg: no, ever "rename" given an pk ID) then there is no potential race condition because of that.

internal/converter/grpc_converter_util.go
```diff
@@ -0,0 +1,371 @@
...
+	"fmt"
+	"github.com/opendatahub-io/model-registry/internal/ml_metadata/proto"
+	"github.com/opendatahub-io/model-registry/internal/model/db"
+	"google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/types/known/structpb"
+	"sync"
+)
+
+var typeNameCache = map[int64]string{}
+var cacheMutex = sync.Mutex{}
```
tarilabs, line 13: not exported (especially the lock) so as far as I understand this is the way "idiomatic for Go" to encapsulate the lock this way

tarilabs (approved):
See https://github.com/opendatahub-io/model-registry/pull/38#issuecomment-1748397964
I have no other questions from me, thanks

lampajr (approved):

PR #39, opened 2023-09-27
implement gRPC rest gateway

### LINKED ISSUE #32: [feature][grpc] Experiment gRPC gateway
UI can't directly do gRPC requests, therefore we should provide a REST proxy  service that can translates REST reqeusts to gRPC ones.

One possible solution could be to autogenerate what is called `gRPC gateway` [1].

None

**Additional context**
This step should be automated in the build Makefile.

[1] https://grpc-ecosystem.github.io/grpc-gateway/

lampajr:
## Description
Fixes https://github.com/opendatahub-io/model-registry/issues/32

Automate the generation of a gRPC rest gateway proxy in front of the actual gRPC server.

Startup the gateway:
```bash
$ ./model-registry gateway
```
or
```bash
$ make run/gateway
```

> **NOTE**: right now I setup a 10s timeout on gRPC gateway, this means that if you start it up and the gRPC is not running yet it will fail and exit after 10s

## How Has This Been Tested?

In order to test this I created a new python script `test/python/test_mlmetadata_rest.py` that aims to create some ArtifactType and an Artifact instance dealing with the REST interface instead of the gRPC one.

Steps to run the tests:
```bash
# in one terminal
$ make run/server

# in another terminal
$ make run/gateway

# in third terminal
$ python test/python/test_mlmetadata_rest.py
```

## Merge criteria:


rareddy:
I there a reason why we want this to be a different process from the main gRPC server? Can not wire up such that it is a single process?

dhirajsb:
@lampajr this looks quite good :+1: 
I'll review it in more detail tomorrow.

lampajr:
> I there a reason why we want this to be a different process from the main gRPC server? Can not wire up such that it is a single process?

Initially because as per my understanding the gRPC gateway needs to dial a connection to the gRPC server (that is why I thought having different processes would have been a better approach) - on the other hand I am evaluating if we can sort this out having a single process as well.

dhirajsb (changes requested):
@lampajr let's hold on to this PR for now while we investigate the envoy proxy that google's ml-metadata server uses. 
@tarilabs noticed that it might be exposing an openapi v3 service, which would be preferable.

lampajr:
Closing as not needed for now.

PR #51, opened 2023-10-04
Define the Model Registry OpenAPI

### LINKED ISSUE #44: Define the Model Registry OpenAPI contract

Need to develop the OpenAPI contract that can be used by UI/DSP/Model Serving teams for Model Registration, Model Searching, Model Retrieval, Model Annotation.

I think we should use a tool like APICurio and build the contract first. Using this we can generate client and server-side stubs that need to be implemented. Here is what I am thinking based on some inspiration from discussions

```
/model-registry
    GET              []RegistreredModel (get all registered models)
    POST            VersionedModel
/search?[name=xxxx]&[id=xxx]&[date=xxx]&[tag=xxx]&[label=xxx]
     GET            []RegistreredModel (get all registered models)
/name
      POST    {VersionedModel}   VersionedModel     Assign a version to a registered model 
/name/version
    GET          VersionedMode
    DELETE   VersionedModel
    PATCH.   {VersionedModel}.  update a model (only label, tags, description)
/stage
    GET.          []stage.       Available stages
    POST        {stage}
    DELETE

// should we define a transition of stages??

RegisteredModel {
    name
    [] vesions
    [] label
}

VesionedModel {
    name
    id
    version
    [] tag ??
    [] label
    stage
    created date
    updated date
}

stage {
   name
   environment urls etc   
}
```

tarilabs:
Resolves #44 

## Description
Implements the OpenAPI as mentioned and discussed in the linked GH issue.

## How Has This Been Tested?
This is a compliant OAS file, generated while using Apicurio studio

## Merge criteria:

- [n/a] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [n/a] The developer has manually tested the changes and verified that the changes work

### DIFF
--- a/api/openapi/modelregistry.yml
+++ b/api/openapi/modelregistry.yml
@@ -1,0 +1,367 @@
+openapi: 3.0.2
+info:
+    title: Model Registry API
+    version: 1.0.0
+    description: 'See also https://github.com/opendatahub-io/model-registry/issues/44#issue-1919724627'
+paths:
+    /environments:
+        summary: Path used to manage the list of environments.
+        description: >-
+            The REST endpoint/path used to list and create zero or more `Environment` entities.  This path
+            contains a `GET` and `POST` operation to perform the list and create tasks, respectively.
+        get:
+            responses:
+                '200':
+                    content:
+                        application/json:
+                            schema:
+                                type: array
+                                items:
+                                    $ref: '#/components/schemas/Environment'
+                    description: Successful response - returns an array of `Environment` entities.
+            operationId: getEnvironments
+            summary: List All Environments
+            description: Gets a list of all `Environment` entities.
+        post:
+            requestBody:
+                description: A new `Environment` to be created.
+                content:
+                    application/json:
+                        schema:
+                            $ref: '#/components/schemas/Environment'
+                required: true
+            responses:
+                '201':
+                    description: Successful response.
+            operationId: createEnvironment
+            summary: Create a Environment
+            description: Creates a new instance of a `Environment`.
+        delete:
+            parameters:
+                -
+                    name: name
+                    description: ''
+                    schema: {}
+                    in: query
+            responses:
+                '200':
+                    description: Successfully deleted `Environment` having `name`.
+            summary: Delete an Environment
+            description: 'Placed here only to avoid subpaths, can be re-drawn.'
+    /search:
+        description: |-
+            TBD: We feel this might be redundant with `models/` and `models/{name}`.
+            We suggest instead to: 
+            1. avoid usage of Query param (that might "explode" as search grows)
+            2. provide search criteria in the body
+        get:
+            responses:
+                '200':
+                    content:
+                        application/json:
+                            schema:
+                                type: array
+                                items:
+                                    $ref: '#/components/schemas/RegisteredModel'
+                    description: Get all `RegisteredModel`s matching the supplied criteria.
+            summary: Search models
+        parameters:
+            -
+                name: name
+                description: ''
+                schema:
+                    type: string
+                in: query
+            -
+                name: id
+                description: ''
+                schema:
+                    type: string
+                in: query
+            -
+                name: tag
+                description: ''
+                schema:
+                    type: string
+                in: query
+            -
+                name: label
+                description: ''
+                schema:
+                    type: string
+                in: query
+    /models:
+        get:
+            responses:
+                '200':
+                    content:
+                        application/json:
+                            schema:
+                                type: array
+                                items:
+                                    $ref: '#/components/schemas/RegisteredModel'
+                    description: List all
+            summary: List all Models
+            description: List all `RegisteredModel` and belonging `VersionedModel`(s).
+        post:
+            requestBody:
+                content:
+                    application/json:
+                        schema:
+                            $ref: '#/components/schemas/VersionedModel'
+                required: true
+            responses:
+                '200':
+                    content:
+                        application/json:
+                            schema:
+                                $ref: '#/components/schemas/VersionedModel'
+                    description: Successfully created a `VersionedModel` for a Registered model.
+            summary: Assign a version to a registered model
+            description: |-
+                If a `RegisteredModel` for `model_name` already exists, this `VersionedModel` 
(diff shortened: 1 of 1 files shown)


tarilabs:
@rareddy as discussed, a static preview can be generated, for example: https://jsfiddle.net/k284ue9r/
If there is a better place where to share+host with the relevant teams, I can make sure to keep that updated.

rareddy:
Thanks @tarilabs The one place I can think of is a way we can push the document on the Model Registry ODH community.

tarilabs:
> The one place I can think of is a way we can push the document on the Model Registry ODH community

@rareddy  once we have a general or overall consensus for the `.yml` file, I can add the `.html` in this repo and then serve it as a GitHub page. wdyt?

tarilabs:
@rareddy html preview of openapi per f692b22 here: [redoc-static.html.zip](https://github.com/opendatahub-io/model-registry/files/12807835/redoc-static.html.zip)

dhirajsb:
@tarilabs this is a good first effort. We need to refactor things a bit so that we can come up with a design pattern for REST resources. 
One of the things we need to change is that for every resource, we need three types, one for create, one for get, and one for update. They should have different fields depending on what's needed/supported for every operation. URL paths also need to follow a pattern with possibly ID based and name based paths. 
Also keeping the REST schema close to the logical metadata model will make it easier to leverage mlmd capabilities later.

api/openapi/modelregistry.yml
```diff
@@ -0,0 +1,367 @@
...
+            responses:
+                '200':
+                    description: Successfully registered a versionedmodel under a registered model.
+            summary: Register a VersionedModel under a given name
+            description: |-
+                TBD: this as a potential alternative or complementary of `/models` POST
+                where the model's name is provided as a Path param instead.
+        parameters:
+            -
+                name: name
```
lampajr, line 204: Without this schema type, the generator complains during the auto-generation phase. Does this sound good to you?

```suggestion
                name: name
                schema:
                    type: string
```

lampajr (changes requested):

rareddy (approved):
lgtm

This is agreed as first draft, to be iterated on

rareddy:
Some Suggestions, for `Environment`
```
{
    "name": "My Environment",
    "uri": "http://localhost:8080"
}
```

I would remove the URI, and add `custom-properties`

In `Artifact` and `RegisteredModel` I would replace metadata with `custom-properties`

on `VersionedModel` 
* `model_uri` should not be there as it is a property of the  `artifact` now.
* can we change `tag` to something like `Deployed Environment`?

> If a RegisteredModel for model_name already exists, this VersionedModel will be assigned to it, else a new RegisteredModel will be created. If a model_name is not provided, a random one will be used.

On thinking again, we do not need a top-level `RegisteredModel` just having `VersionedModel` might suffice. In the backend, we just need to collate the labels onto the `Context` object that represents the collection which is "registeredmodel". then we can have GET/POST on `/models/{name}` that removes the ambiguity with POST call

`/search` seems verbose with `/models`, if we move the query parameters on to `/models` then we can get rid of `/search`

define `Operation ID` on operations, which will generate consistent operation names if we generate code across different languages.

> Some fields will not be updated even if supplied, such as `model_uri`, `id`, `version`.

we should probably not ignore silently but raise a 400 saying bad request if present and values are different from what is in the source.

I did not quite understood the concatenation of name+version as ID you were mentioning but probably should not doit but rather use the MLMD to figure out how we can arrange the data

dhirajsb:
@andrewballantyne this is a rough draft, and is going to undergo changes.

dhirajsb:
@rareddy there are a lots of things we must improve in this API. I'm working on refactoring it and creating a new PR. I was testing the CPP server API today, so it'll take me one more day. 

I do have a question about environment. I know you introduced this to mimic MLFlow environments, but what does it map to in ODH? Is there an ODH component/k8s resource it would map to?

dhirajsb (approved):
/lgtm for now, the API will be refactored based on more detailed REST resource modeling and analysis

tarilabs:
@rareddy 
> On thinking again, we do not need a top-level `RegisteredModel` just having `VersionedModel` might suffice

Doing this would remove the ability to have metadata which pertains to a given model, more than a specific version.
For example if `RegisteredModel.name == "Pricing model"` a collection of metadata could be `{ "ownership" : "marketing" }` regardless of version(s).

Personally, I liked the idea of having a RegisteredModel having multiple VersionedModel, as that somehow aligned analogously to containerimage/versions or repo/release.

The thing which was clarified from #44 is that we have the ability to create a RegisteredModel automatically if it was not already existings, as you ask via API call to MR to index a VersionedModel.

tarilabs:
@rareddy 

> define `Operation ID` on operations, which will generate consistent operation names if we generate code across different languages.

~~Sorry, I haven't understood this one. Could you clarify possibly with an example, please?~~

EDIT: nevermind, @lampajr pointed me in the right direction, I believe you meant this: https://github.com/opendatahub-io/model-registry/issues/44#issuecomment-1741943369

I agree, we should add them as needed as we work on the core/service layer

tarilabs:
@rareddy 

> I did not quite understood the concatenation of name+version as ID you were mentioning but probably should not doit but rather use the MLMD to figure out how we can arrange the data

I believe that discussion was not related to this PR.
We needed a way to uniquely populate mlmd's Context.name once we translate this domain model inside mlmd, and I believe we converged on using `parent_context_id:version`.

tarilabs:
@dhirajsb 

> One of the things we need to change is that for every resource, we need three types, one for create, one for get, and one for update. They should have different fields depending on what's needed/supported for every operation. URL paths also need to follow a pattern with possibly ID based and name based paths.

As discussed, it depends if we want to have a pure REST approach, or concentrate on the bare-minimal operations we need out of the gate. My understanding from #44 was more on the latter.

tarilabs:
@dhirajsb 

> Also keeping the REST schema close to the logical metadata model will make it easier to leverage mlmd capabilities later.

My understanding from https://github.com/opendatahub-io/model-registry/issues/44#issuecomment-1741943369 is that we need to provide a MR focused domain model the API operates on.

lampajr (approved):

PR #64, opened 2023-10-06
Extend GHA to cover for Docker build

### LINKED ISSUE #50: Extend GHA to cover for Docker build
Consider extending GHA to cover testing of the Docker image build which was introduced by https://github.com/opendatahub-io/model-registry/pull/42#event-10529307219

tarilabs:
Resolves #50 

## How Has This Been Tested?
Will appear in Actions tab.
For instance see here: https://github.com/opendatahub-io/model-registry/actions/runs/6432552631/job/17467707214?pr=64#step:3:154

## Merge criteria:

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -1,0 +1,15 @@
+name: Test Container Image build
+
+on:
+  push:
+    branches: [ 'main' ]
+  pull_request:
+    branches: [ '*' ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build the Docker image
+      run: docker build . --file Dockerfile --tag model-registry:$(date +%s)


lampajr (approved):

.github/workflows/container-image.yml
```diff
@@ -0,0 +1,15 @@
+name: Test Container Image build
+
+on:
+  push:
+    branches: [ 'main' ]
+  pull_request:
+    branches: [ '*' ]
```
antisaling, line 7: We could use something like [path exclusion](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths) to avoid running this needlessly.

antisaling (approved):
lgtm, see comment inline.

.github/workflows/container-image.yml
tarilabs: I see your point, but since the builder pattern is used inside the Dockerfile, we would need to check the build also works with most code-change related PRs anyway.

I concur we might skip this for PR focusing on documentation, etc. 👍

PR #66, opened 2023-10-11
Refactor model registry openapi definition.

### LINKED ISSUE #67: Refactor openapi defintion to meet RESTful API standards
Model registry openapi definition should meet standard REST API development guidelines as outlined by Architecture Council members. 

OpenAPI should follow REST patterns, and should closely match Kubernetes resource APIs, which is the API being used by other ODH components for controllers, CRDs, etc. 

None.

**Additional context**
The REST API should also leverage boilerplate patterns developed in managed application services over several years to avoid having to relearn those lessons and reinvent the wheel.

dhirajsb:
## Description

Refactor model registry to:
* Use request/response types
* Use lists with paging options
* Add support for orderBy param
* Fix resource URLs to fit RESTful API scheme
* Use reusable response types, and base json schema types to compose messages to minimize duplication of common properties

Resolves #67 

## How Has This Been Tested?

WIP

## Merge criteria:

- [ ] The commits and have meaningful messages; the author will squash them [after approval](https://github.com/opendatahub-io/opendatahub-community/blob/main/contributor-cheatsheet.md#:~:text=Usually%20this%20is%20done%20in%20last%20phase%20of%20a%20PR%20revision) or will ask to merge with squash.


dhirajsb:
Need to add openapi generator to Makefile to make sure the API can be compiled to generate expected type model in Golang for implementing REST proxy server.

lampajr:
> Need to add openapi generator to Makefile to make sure the API can be compiled to generate expected type model in Golang for implementing REST proxy server.

May I suggest to use [oapi-codegen](https://github.com/deepmap/oapi-codegen) to generate type models? 

```bash
oapi-codegen -package <package-name> -generate types api/openapi/model-registry.yaml >> internal/model/models.gen.go
```

It seems pretty good in autogenerating stuff from the openapi spec for Go projects, furthermore we can also generate server stub for different Go frameworks (like `gin`) by adding additional `-generate` props [1].

[1] https://github.com/deepmap/oapi-codegen#using-oapi-codegen

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,827 @@
...
+        description: Updated `RegisteredModel` information.
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/RegisteredModelUpdate'
+        required: true
+      tags:
+        - ModelRegistryService
+      responses:
+        '202':
```
rareddy: 202? any reason behind this not being 200?

dhirajsb:
> May I suggest to use [oapi-codegen](https://github.com/deepmap/oapi-codegen) to generate type models?

Do we have any other odh/openshift projects that have used oapi-codegen? It would be nice to know if there are any known issues/tricks to use it.

lampajr:
> Do we have any other odh/openshift projects that have used oapi-codegen? It would be nice to know if there are any known issues/tricks to use it.

I couldn't find any usage so far but I will investigate further. 

FWIW I tried it in the Go api PR for the types generation only.

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,827 @@
...
+              type: array
+              items:
+                $ref: '#/components/schemas/Artifact'
+        - $ref: '#/components/schemas/BaseResourceList'
+    OrderByField:
+      description: 'Supported fields for ordering result entities. '
+      enum:
+        - CREATE_TIME
+        - LAST_UPDATE_TIME
+        - ID
```
rareddy: can we add by name, author name here?

rareddy (commented):
@dhirajsb good job! Can you explain the difference between

`/api/model_registry/v1alpha1/model_artifact` Vs `/api/model_registry/v1alpha1/model_versions/{modelversionId}/artifacts`

similarly 

`/api/model_registry/v1alpha1/model_version` vs `/api/model_registry/v1alpha1/registered_models/{registeredmodelId}/versions`

artifacts I can see they can exist across the models maybe there is a requirement to fetch them directly, but model versions do not have that , 

I would have thought in true rest fashion I would see
GET/POST `/api/model_registry/v1alpha1/registered_models/{registeredmodelId}/versions/{modelversionId}/artifact`  any reason for this to be not there?

I did not see any `deploy stage` specific attributes on the model version.

The Execution stuff not being used, any purpose for that?

I would like here your interpretation of relation with RegisteredModel and ModelVersion to clear some of questions above

api/openapi/model-registry.yaml
dhirajsb: Good catch. Fixed by changing it to 201 for POST.

api/openapi/model-registry.yaml
dhirajsb: Unfortunately, those are the only supported `orderBy` fields in the `ml-metadata` API. 
I am guessing that the idea is that if the user is looking for specific values in text fields like `name`, they should probably use them in a search query to filter results.

dhirajsb:
> @dhirajsb good job! Can you explain the difference between
> 
> `/api/model_registry/v1alpha1/model_artifact` Vs `/api/model_registry/v1alpha1/model_versions/{modelversionId}/artifacts`
> 

`ModelArtifacts` exist independent of model versions like you already guessed later. So a user could create them right away from an experiment run, and then associate it with a specific `ModelVersion` later using the second url. Or, they can also directly create the `ModelArtifact` and associate it with a `ModelVersion` using the second url if it's from a downloaded model like one from huggingface. 

> similarly
> 
> `/api/model_registry/v1alpha1/model_version` vs `/api/model_registry/v1alpha1/registered_models/{registeredmodelId}/versions`
> 
The `ModelVersionCreate` message has a `registeredModelID` property. So a user could just use the first url to create versions directly for a `RegisteredModel`, or use the second url to provide the id in the path. It's just a convenience, and trying to make resources and their urls more boilerplate. Internally the service implementation is going to be the same. 
> artifacts I can see they can exist across the models maybe there is a requirement to fetch them directly, but model versions do not have that ,
> 
If a client has a `ModelVersion` id, it can retrieve it directly from the resource's dedicated url like `.../model_versions/`.  Client can also search for a `ModelVersion` using an `externalID`, which could be a git tag/commit from either a huggingface or a Jupyter notebook git project. 
The urls for children resources can also be used for more advanced use cases later like searching by timestamps to pull all new `ModelVersions` created after the last time model serving reconcile loop ran. 

> I would have thought in true rest fashion I would see GET/POST `/api/model_registry/v1alpha1/registered_models/{registeredmodelId}/versions/{modelversionId}/artifact` any reason for this to be not there?
> 
I am avoiding nesting resources more than 2 levels deep. If there is a child resource, it should have it's own root url, and then sub paths for resources directly associated with it. It creates a boilerplate and makes it easier to manage relationships. 
Later when we start supporting lineage searches, like going back from artifacts to contexts, executions to contexts and artifacts, etc. the forward and reverse relationships will become hard to follow if we nest resources deeply. 
> I did not see any `deploy stage` specific attributes on the model version.
> 
I added properties from model serving CRD to `ModelArtifact`. 
> The Execution stuff not being used, any purpose for that?

As we discussed earlier, we need to discuss with model serving how to model `DeploymentEnvironments`, which might affect how the relationship between environments and versions is handled and where the `Deploy` execution finally ends up being created. 
I'll start that conversation tomorrow on model serving channel and include the MR team. 

[edit: formatting]

dhirajsb:
I'll use `[oapi-codegen](https://github.com/deepmap/oapi-codegen)` and cleanup any potential model generation specific issues tomorrow. Also, we'll run current execution mapping idea by model-serving folks since they are the domain SMEs. And they can also guide us on how to model deployment environments.

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1041 @@
...
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+      operationId: findModelVersion
+      summary: Get a ModelVersion that matches search parameters.
+      description: Gets the details of a single instance of a `ModelVersion` that matches search parameters.
+    parameters:
+      -
+        $ref: '#/components/parameters/externalID'
```
lampajr: Regarding `/api/model_registry/v1alpha1/model_version` is there any reason why don't we provide capabilities to find Model Version by `name` and `version`? I think this is something that could be useful (or even needed)

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1041 @@
...
+      schema:
+        type: string
+      in: query
+      required: false
+    orderBy:
+      style: form
+      explode: true
+      examples:
+        orderBy:
+          value: name asc
```
tarilabs:
```suggestion
          value: "CREATE_TIME"
```

We didn't get also if ` asc` or ` desc` postfix is supported in enum? 🤔

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1041 @@
...
+      default: UNKNOWN
+      enum:
+        - UNKNOWN
+        - PENDING
+        - LIVE
+        - MARKED_FOR_DELETION
+        - DELETED
+        - ABANDONED
+        - REFERENCE
+      type: string
```
tarilabs: Just to better documentation, is this coming from some ml-metadata or other requirements, or is this something which we are introducing in MR-domain with this openAPI iteration?

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1041 @@
...
+      type: object
+      allOf:
+        -
+          type: object
+          properties:
+            lastKnownState:
+              $ref: '#/components/schemas/ExecutionState'
+        -
+          $ref: '#/components/schemas/BaseResourceUpdate'
+    MetadataValue:
```
tarilabs, line 1165: We understand this is derived from ml-metatdata protobuf, but shouldn't this be just a generic object, and for some default mapping to be applied?
ie: avoid for the end-user of the MR to know about this implementation details

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1041 @@
...
+            name:
+              description: |-
+                The client provided name of the artifact. This field is optional. If set,
+                it must be unique among all the artifacts of the same artifact type within
+                a database instance and cannot be changed once set.
+              type: string
+    BaseResourceUpdate:
+      type: object
+      properties:
+        customProperties:
```
tarilabs, line 1245:
```suggestion
        metadata:
```

open question, why not calling it directly metadata?

api/openapi/model-registry.yaml
tarilabs: Possibly can respond a 400 response if supplied name and forgot to supply version (name-and-version not found already have 404)

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1041 @@
...
+      name: id
+      description: The ID of resource.
+      schema:
+        type: string
+      in: path
+      required: true
+    name:
+      examples:
+        name:
+          value: artifact-name
```
tarilabs:
```suggestion
          value: entity-name
```

since we understand this is a generic params/example, this would avoid confusion in generated doc such as:
Screenshot 2023-10-12 at 11 26 15

as in, while speaking of registered model, the example speaks of artifact

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1041 @@
...
+          description: Maximum number of resources to return in the result.
+          type: integer
+        size:
+          format: int32
+          description: Number of items in result list.
+          type: integer
+    Artifact:
+      oneOf:
+        -
+          type: object
```
tarilabs: Was this intentional?
This creates an optional "free form" JSON in many POSTs, such as:
image

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1041 @@
...
+          $ref: '#/components/schemas/BaseResourceCreate'
+        -
+          $ref: '#/components/schemas/RegisteredModelUpdate'
+    RegisteredModelUpdate:
+      description: 'A registered model in model registry. A registered model has OdhModelVersion children. '
+      allOf:
+        -
+          $ref: '#/components/schemas/BaseResourceUpdate'
+    ModelVersion:
+      description: 'Represents a collection of artifacts and executions for a version of a RegisteredModel. '
```
tarilabs:
```suggestion
      description: 'Represents a ModelVersion associated to a RegisteredModel.'
```

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1041 @@
...
+      allOf:
+        -
+          $ref: '#/components/schemas/ModelVersionCreate'
+        -
+          $ref: '#/components/schemas/BaseResource'
+    ModelVersionCreate:
+      description: 'Represents a collection of artifacts and executions for a version of a RegisteredModel. '
```
tarilabs:
```suggestion
      description: 'Represents a ModelVersion associated to a RegisteredModel. '
```

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1041 @@
...
+        -
+          $ref: '#/components/schemas/BaseResourceCreate'
+        -
+          $ref: '#/components/schemas/ModelVersionUpdate'
+      properties:
+        registeredModelID:
+          description: ID of the `RegisteredModel` to which this version belongs.
+          type: string
+    ModelVersionUpdate:
+      description: 'Represents a collection of artifacts and executions for a version of a RegisteredModel. '
```
tarilabs:
```suggestion
      description: 'Represents a ModelVersion associated to a RegisteredModel. '
```

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1041 @@
...
+        -
+          $ref: '#/components/schemas/BaseResourceCreate'
+        -
+          $ref: '#/components/schemas/RegisteredModelUpdate'
+    RegisteredModelUpdate:
+      description: 'A registered model in model registry. A registered model has OdhModelVersion children. '
+      allOf:
+        -
+          $ref: '#/components/schemas/BaseResourceUpdate'
+    ModelVersion:
```
tarilabs, line 1094: Possible a "human readable", optional `version` property is needed?

tarilabs (commented):
IIUC, to index an Artifact on MR using this RESTful approach requires in turn:
1. create a RegisteredModel,
2. take id from previous step, create a ModelVersion,
3. take id from previous step, create a ModelArtifact associated with Model Version

would it be helpful to consider a non-RESTful utility endpoint, which at least collapses step 2 and repeated step 3s (associate many files) in a single call?
Or to do this directly inside operation `createModelVersion` ?

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1041 @@
...
+        storagePath:
+          description: Path for model in storage provided by `storageKey`.
+          type: string
+        modelFormatVersion:
+          description: 'Version of the model format. '
+          type: string
+        serviceAccountName:
+          description: 'Name of the service account with storage secret. '
+          type: string
+    ModelArtifactCreate:
```
lampajr, line 1313: For storage artifacts that are not s3-based, it could be helpful to have a generic `uri` property at ModelArtifact level.

tarilabs:
Noted with @lampajr as we were experimenting with oapi-codegen, it seems to have difficulties in normalizing properties in "inheritance plus additional props" case.

That is, seems is not generating go struct fields for types such as:

```yaml
description: 'An ML model artifact. '
type: object
allOf:
    -
        $ref: '#/components/schemas/BaseArtifactUpdate'
properties:
    modelFormatName:
        description: 'Name of the model format. '
        type: string
    runtime:
        description: 'Model runtime. '
        type: string
    storageKey:
        description: 'Storage secret name. '
        type: string
    storagePath:
        description: Path for model in storage provided by `storageKey`.
        type: string
    modelFormatVersion:
        description: 'Version of the model format. '
        type: string
    serviceAccountName:
        description: 'Name of the service account with storage secret. '
        type: string
```

so `modelFormatName`, `runtime`, `storageKey` are not codegen rendered in the target Go struct.

Based on Swagger/OAS examples from the spec, an easy fix seems to be refactoring by adopting an anonymous object as part of the `allOf`, for example:

```yaml
description: 'An ML model artifact. '
type: object
allOf:
    - $ref: '#/components/schemas/BaseArtifactUpdate'
    - type: object
      properties:
          modelFormatName:
              description: 'Name of the model format. '
              type: string
          runtime:
              description: 'Model runtime. '
              type: string
# ...
```

rareddy:
>I am avoiding nesting resources more than 2 levels deep. 

Can we avoid this altogether, and leave it level 1? IMO it is confusing if we do this for some not all the way. This will reduce the API footprint as you already have the root URL for the 2nd levels anyway.

api/openapi/model-registry.yaml
rareddy: IMO, `custom properties` convey the intent more clearly as they are arbitrary data. `metadata` without a schema attached may be vague.

api/openapi/model-registry.yaml
rareddy: isn't that coming from BaseArtifact?

lampajr: you are right, I completely missed that :/

tarilabs: we might have been misled because of the problem discussed in https://github.com/opendatahub-io/model-registry/pull/66#issuecomment-1759630182

dhirajsb: There is a `uri` property in the `BaseArtifactUpdate` type.

api/openapi/model-registry.yaml
dhirajsb: The `name` property of the `ModelVersion` is the version label.

api/openapi/model-registry.yaml
dhirajsb: This was a bug introduced because of ApiCurio studio. Removed the empty object.

api/openapi/model-registry.yaml
dhirajsb: Good catch, I had meant to add the `name` query parameter. Eventually there would be a generic SQL like `query` parameter where `name` could be specified. But, we'll create one for just the `name` that will be implemented using `ILIKE` operator.

api/openapi/model-registry.yaml
dhirajsb: Yup, meant to add another query parameter for order direction.

api/openapi/model-registry.yaml
dhirajsb: Instead of re-inventing the wheel for artifact state, I lifted this from mlmd as is. As we develop and use the API more and see real world use cases, we can either keep this if it suites our needs or add a new property with less/more enumeration states. wdyt?

api/openapi/model-registry.yaml
dhirajsb: Ml-metadata chose to do it like this for a reason. I'm assuming that having a value type helps validates it on the client side, e.g. a client cannot accidentally send a typo in a bool. And the consumer of these user provided properties might also find it useful to get type information.

api/openapi/model-registry.yaml
dhirajsb: These are meant for user provided custom properties.

dhirajsb:
@tarilabs @lampajr I'm seeing the same issue with `Base*` types. It's possible it could be an oapi-codegen issue. I'll test it with openapi-generator tomorrow.

dhirajsb:
> Can we avoid this altogether, and leave it level 1? IMO it is confusing if we do this for some not all the way. This will reduce the API footprint as you already have the root URL for the 2nd levels anyway.

I agree that the duplicate endpoints seem to be cluttering the API at first glance. But having one level sub-path for children for every entity shows all the other entities that can be reached from that entity. Sort of like a recursive graph node visitor pattern. 

So, the REST API self describes a boilerplate way to traverse the graph, without having to read any documentation on all the ways the entities are related to each other. 

Once a user becomes familiar with the relationships, they may just use the root entity paths directly all the time.

api/openapi/model-registry.yaml
tarilabs: > wdyt?

Indeed as originally mentioned I'd add for better documentation where this is coming from, something ~like:

```suggestion
    ArtifactState:
      description: |4-
           - PENDING: A state indicating that the artifact may exist.
           - LIVE: A state indicating that the artifact should exist, unless something
          external to the system deletes it.
           - MARKED_FOR_DELETION: A state indicating that the artifact should be deleted.
           - DELETED: A state indicating that the artifact has been deleted.
           - ABANDONED: A state indicating that the artifact has been abandoned, which may be
          due to a failed or cancelled execution.
           - REFERENCE: A state indicating that the artifact is a reference artifact. At
          execution start time, the orchestrator produces an output artifact for
          each output key with state PENDING. However, for an intermediate
          artifact, this first artifact's state will be REFERENCE. Intermediate
          artifacts emitted during a component's execution will copy the REFERENCE
          artifact's attributes. At the end of an execution, the artifact state
          should remain REFERENCE instead of being changed to LIVE.
           See also: ml-metadata Artifact.State
      default: UNKNOWN
      enum:
        - UNKNOWN
        - PENDING
        - LIVE
        - MARKED_FOR_DELETION
        - DELETED
        - ABANDONED
        - REFERENCE
      type: string
```

tarilabs:
In the hope it could also help, here is a draft sequence diagram for the "indexing a model (artifact) inside of MR as a downstream of a notebook/pipeline execution" in the way I intended with this PR changes:

```mermaid
sequenceDiagram
    actor DS as Data Scientist<br/>/pipeline
    participant S as Platform Storage<br/>(S3)
    participant MR as Model Registry
    DS-->>DS: <br/><br/>Model training completes
    DS->>S: Persist model
    S-->>DS: s3 uri
    DS->>MR: createRegisteredModel
    MR-->>DS: RegisteredModel id
    DS->>MR: createModelVersion (using RegisteredModel id)
    MR-->>DS: ModelVersion id
    DS->>MR: createModelVersionArtifact (using ModelVersion id and uri)
    MR-->>DS: ModelArtifact id
```

Doing so I noticed if using the operation `createModelArtifact` (instead of `createModelVersionArtifact`) seems to me there was not way to associate the ModelArtifact ex-post its creation to the ModelVersion? 🤔 (or at least I might have missed how)

Edit: Or alternatively when considering the eventual Python/GoLang/"client" binding SDK more precisely:

```mermaid
sequenceDiagram
    actor DS as Data Scientist<br/>/pipeline
    participant S as Platform Storage<br/>(S3)
    participant SDK as SDK/bindings for MR
    participant MR as Model Registry
    DS-->>DS: <br/><br/>Model training completes
    DS->>S: Persist model
    S-->>DS: s3 uri
    DS->>SDK: Intent to index ML model on MR
    SDK->>MR: createRegisteredModel
    MR-->>SDK: RegisteredModel id
    SDK->>MR: createModelVersion (using RegisteredModel id)
    MR-->>SDK: ModelVersion id
    SDK->>MR: createModelVersionArtifact (using ModelVersion id and uri)
    MR-->>SDK: ModelArtifact id
```

rareddy:
> Once a user becomes familiar with the relationships, they may just use the root entity paths directly all the time.

I am not against one way or the other. I am trying to understand limiting it to the 2nd level of depth, only thing I can think of is long URLs which can get hairy in terms of readability.

rareddy:
I do not see any model `deployment state` enumerations like `new`, `ready for deployment`, `deploying`, `deployed`,`failed to deploy`,  `archived`, `deleted`. Can we define these in the API. Also, we should also consider that a given model version could have been deployed into more than one deployment environment like a model could have been deployed in `testing` as well as in `production` at the same time in different environments.

rareddy:
>Edit: Or alternatively when considering the eventual Python/GoLang/"client" binding SDK more precisely:

That is way I think about it too, the REST API /Go API layer is more of a one-to-one with MLMD API where the type-id is defined and logical entity names are used (loosely speaking), then we should provide higher-level utility functions in Go & Python to hide any common use cases.

dhirajsb:
I fixed the missing properties for `BaseArtifactUpdate` and `ModelArtifactUpdate` introduced by some weird UI issue in Apicurio-Studio. 
I used openapi-generator on Friday and the results are better than oapi-codegen. It also produces types in separate files instead of one giant go file. So, imho we should use the tooling we have experience with in other Red Hat projects. 
There were some weird code vetting issues from the generated client and service code, which I'm going to fix tomorrow.

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1070 @@
...
+              format: int64
+              description: Output only. Create time of the resource in millisecond since epoch.
+              type: string
+              readOnly: true
+            lastUpdateTimeSinceEpoch:
+              format: int64
+              description: |-
+                Output only. Last update time of the resource since epoch in millisecond
+                since epoch.
+              type: string
```
tarilabs, line 1227: Open question: any reason why these base properties are JSON's `string` serializing a number?
Isn't it better to keep them `integer` in the [JSON Schema](https://json-schema.org/understanding-json-schema/reference/numeric#integer)? 🤔

lampajr:
After some discussion on the service layer api we struggled understanding what is the real meaning/usage of the `name` parameter for ModelArtifact and ModelVersion entities.

This was my understanding:
* The `name` should be a user-provided string which should be unique in a specific context and not in the whole database
  * ModelVersion name should be unique among all model versions belonging to the same RegisteredModel
  * ModelArtifact name should be unique among all model artifacts under a specific ModelVersion (if associated)
  * The service layer has guarantee the uniqueness in the correct contexts correctly mapping that user-defined name to the internal mlmd name column, e.g, using <registeredModelId>:<versionName> for ModelVersion entities as we already discussed last week
* I would hide the internal mlmd name column to the end user, as I don't think it is something they should care about (this is internal logic)

Based on the openaAPI spec, mainly on `find<entity>` operations, the `name` parameter seems exactly the name column in the mlmd database but _I am not expecting the user should know how we are internally managing the uniqueness constraints_, therefore here is how I would have imagined those two endpoints parameters:
* `findModelVersion` : find by `<externalId>` or `<registeredModelName, versionName>`
 * `findModelArtifact`: find by `<externalId>` or `<modelVersionId, name>` or `<registeredModelName, modelVersionName, name>`

I  think this is a crucial aspect to settle down, happy to discuss in a mtg if preferrable.

api/openapi/model-registry.yaml
dhirajsb: Added the suggested description change.

api/openapi/model-registry.yaml
dhirajsb: Changed to `ModelVersion belonging to a RegisteredModel`.

api/openapi/model-registry.yaml
dhirajsb: The problem is that there is no native int64 in Javascript. This mapping is also what we did in rhoc and rhosak.

dhirajsb:
As discussed, artifact names will be made unique in the service layer by adding a uuid to names internally before creating the underlying artifact instance in mlmd server. This is to avoid running into the unique constraint that requires names to be unique among all instances of any resource type.

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1713 @@
...
+          $ref: '#/components/schemas/BaseResourceCreate'
+        -
+          $ref: '#/components/schemas/InferenceServiceUpdate'
+      properties:
+        registeredModelId:
+          description: ID of the `RegisteredModel` to serve.
+          type: string
+        servingEnvironmentId:
+          description: ID of the parent `ServingEnvironment` for this `InferenceService` entity.
+          type: string
```
tarilabs: not necessary in this PR (since is for _Serving_) but here I believe is missing the "anonymous object" ? 🤔 

```suggestion
      allOf:
        -
          $ref: '#/components/schemas/BaseResourceCreate'
        -
          $ref: '#/components/schemas/InferenceServiceUpdate'
        -
          type: object
          properties:
            registeredModelId:
              description: ID of the `RegisteredModel` to serve.
              type: string
            servingEnvironmentId:
              description: ID of the parent `ServingEnvironment` for this `InferenceService` entity.
              type: string
```

api/openapi/model-registry.yaml
```diff
@@ -0,0 +1,1713 @@
...
+        from a `RegisteredModel` created by Model Serving.
+      allOf:
+        -
+          $ref: '#/components/schemas/BaseResourceUpdate'
+      properties:
+        modelVersionId:
+          description: >-
+            ID of the `ModelVersion` to serve. If it's unspecified, then the latest `ModelVersion`
+            by creation order will be served.
+          type: string
```
tarilabs: Analogous comment, for "anonymous object" in combination with `allOf:`

tarilabs (dismissed):
I'd prefer to merge this so we ease the work on the other PRs (https://github.com/opendatahub-io/model-registry/pull/68)

Serving-related aspects imho can be dealt with in another iteration.

lampajr (dismissed):
LGTM as first iteration, thanks @dhirajsb !

api/openapi/model-registry.yaml
dhirajsb: Fixed.

api/openapi/model-registry.yaml
dhirajsb: Fixed.

tarilabs (approved):
I'd prefer to merge this so we ease the work on the other PRs (https://github.com/opendatahub-io/model-registry/pull/68)

Serving-related aspects imho can be dealt with in another iteration.

api/openapi/model-registry.yaml
tarilabs: 👍

api/openapi/model-registry.yaml
tarilabs: 👍

dhirajsb:
> I'd prefer to merge this so we ease the work on the other PRs (https://github.com/opendatahub-io/model-registry/pull/68)

> Serving-related aspects imho can be dealt with in another iteration.

I had already mentioned that in the previous meeting, in case you missed it. This is an initial version of the model-serving resources. I don't think we can make it more minimal. The design was also agreed upon in a previous scrum that showed how it can handle various versioning and serving use cases. 

I think the only changes we might have to make further are using different names if model-serving wants to, and adding more properties if they wish to expose any metadata from model-serving k8s resources.

tarilabs:
> I had already mentioned that in the previous meeting, in case you missed it.

Sure, I was recording what I also mentioned in the previous meeting in the PR for the records.

I think you had already approvals as discussed and as discussed you may proceed to merge this PR first, as you wanted.

dhirajsb:
@andrewballantyne fyi, this pushes a refactored version of the model-registry OpenAPI.

andrewballantyne:
@dhirajsb excellent... we have not started digging into it -- we await the first pass of UX before we start investigating how the API will back it. Is there an easy / reliable way we can see the swagger page for whatever the state of API is at a given point of time?

tarilabs:
@andrewballantyne 
> Is there an easy / reliable way we can see the swagger page for whatever the state of API is at a given point of time?

I think one way is using this link: https://editor.swagger.io/?url=https://raw.githubusercontent.com/opendatahub-io/model-registry/main/api/openapi/model-registry.yaml

i.e.: combine swagger UI generators or similar to github raw link which supports CORS for the link on main branch of the OpenAPI file.

PR #68, opened 2023-10-11
Implement MR service layer

### LINKED ISSUE #58: Create a Go API for Model Serving
For the Model Controller of the ODH Model Serving Operator to talk to the Model Registry  an API is required

Define a Go API interface for Model Registry for
* Listing registered models
* Listing models by search criteria - such as by deployment, by stage, by name & version, by id.
* Add deployment status to Modelversion (need to figure out what metadata with the serving team). Consider environment details here

lampajr:
Fixes https://github.com/opendatahub-io/model-registry/issues/58

## Description

This PR aims to provide a first implementation of a Go-based model registry service layer, providing a minimal API based on the current `api/openapi/model-registry.yaml`.

Main aspects:
- `/internal/core/mapper/mlmd_mapper.go` contains a simple Internal <--> MLMD model mapping.
- `/internal/core/core.go` implements the core service layer named `ModelRegistryService` which provides a minimal API to:
  - Create/Get a `RegisteredModel`
  - Create/Get a `ModelVersion`
  - Create/Get a `ModelArtifact`
  - This API was strictly inspired by the OpenAPI spec
- `/internal/core/core_test.go` contains some tests we were able to setup showing how the service could be used.
  - These tests have been implemented using TestContainer based testing approach, therefore running those tests against a mlmd test container.
- `/test/bdd/` contains BDD test using cucumber 

As an example for the Cucumber approach we implemented a simple User Story, here the result:

image

## How Has This Been Tested?

This has been tested running the newly created tests (i.e., `core_test.go`):

```bash
make test
```

## Merge criteria:


examples/mr_service_example.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	v1Json, _ := json.MarshalIndent(v1, "", "  ")
+	fmt.Printf("Getting model version: %v\n", string(v1Json))
+
+	// [END]: demo showcasing model registration with multiple versions
+
+	glog.Info("shutdown!")
+}
+
+func regiserModelVersion(service core.ModelRegistryApi, name string, version string, author string, uri string, format string) {
```
antisaling: typo:

```suggestion
func registerModelVersion(service core.ModelRegistryApi, name string, version string, author string, uri string, format string) {
```

README.md
```diff
@@ -83,3 +83,13 @@ And `<hostname>` and `<port>` are the local ip and port to use to expose the con
 The server listens on `localhost` by default, hence the `-n 0.0.0.0` option allows the server port to be exposed. 
 
 Once the server has started, test clients and playground can be used as described in the above sections. 
+
+## Domain Sepcific Model
+
+The internal domain specific model is auto-generated from the OpenAPI specification, running the following command:
+
+```bash
+$ oapi-codegen -package registry -generate types api/openapi/modelregistry.yml >> internal/model/registry/registry.gen.go
```
antisaling: Wouldn't it make more sense to

```suggestion
make gen/oapi_models
```

here?

Also, I prefer omitting the `$` as we're not providing the command output, so the user can simply copy the command from the box as is.

README.md
```diff
@@ -83,3 +83,13 @@ And `<hostname>` and `<port>` are the local ip and port to use to expose the con
...
+```
+
+> **Note**: this step is automated in the Makefile `generate` step.
```
antisaling: Also, there's no `generate` step, only `gen`, so it might generate some confusion, especially as we're only documenting the `build` step.

examples/mr_service_example.go
```diff
@@ -0,0 +1,138 @@
...
+		case logging.LevelDebug:
+			msg = fmt.Sprintf("DEBUG :%v", msg)
+		case logging.LevelInfo:
+			msg = fmt.Sprintf("INFO :%v", msg)
+		case logging.LevelWarn:
+			msg = fmt.Sprintf("WARN :%v", msg)
+		case logging.LevelError:
+			msg = fmt.Sprintf("ERROR :%v", msg)
+		default:
+			panic(fmt.Sprintf("unknown level %v", lvl))
```
antisaling: We should probably use a logging library here, though I see that neither `log` nor `glog` provide `Debug` logging. I found the [`logrus` package](https://pkg.go.dev/github.com/sirupsen/logrus) while looking for something but IDK whether it's worth the hassle to adopt it (I suppose we'd have to change logging in other parts of the code too) but [`go-grpc-middleware` supports that too](https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus) ([example usage](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/main/interceptors/logging/examples/logrus/example_test.go)).

examples/mr_service_features_test.go
```diff
@@ -0,0 +1,153 @@
...
+		context.Background(),
+		mlmdAddr,
+		grpc.WithReturnConnectionError(),
+		grpc.WithBlock(),
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	)
+	if err != nil {
+		log.Fatalf("Error dialing connection to mlmd server %s: %v", mlmdAddr, err)
+		return nil, err
+	}
```
antisaling: do we want to setup the connection inside the test? I suppose this should be part of the setup function.

antisaling (commented):
Just a preliminary review as I was skimming through the code.

Edit: put remarks on separate comments so we can discuss further.

antisaling:
I'd separate the functionality of Mapper as the two uses are pretty distinct, i.e. MLMD -> Internal conversion would be separate from Internal -> MLMD.

README.md
lampajr: +1 just a typo, thanks for spotting it

examples/mr_service_example.go
lampajr: I definitely agree on switching to use some ad-hoc logging libraries, for now I just used the same approach that was used in other files. I wouldn't pay too much attention on this as this is just for the example.

Anyway `go-grpc-middleware` looks pretty much interesting, will give it a look :)

examples/mr_service_features_test.go
lampajr: from my understanding this is part of `iHaveAConnectionToMR` which is a precondition of the Cucumber test, therefore it can be seen exactly as a setup here. @tarilabs is my understanding correct?

tarilabs: I would expect is not the responsibility of the "core/service" layer to setup the connection to grpc opaquely, so yes like the example already build a connection and passes over the the core-layer, same happens in testing.

This is what I understood we wanted to create with the core-layer.
For analogy, the "DAL" worked in the same way (it did not setup the gorm/db connection opaquely).

I am always open to consider refactorings if we change our mind.

tarilabs (dismissed):
fwiw I've made use of this core/service layer while contributing TestContainers and cucumber testing. The core/service layer works for me. While I was not part of reviewer (since already contributed commits here) now recording my approval since github switched me as "commenter only" as participating in the discussion, so tidying that up.

lampajr:
As agreed we are going to rebase this PR from the [OpenAPI refactoring one](https://github.com/opendatahub-io/model-registry/pull/66) - this is going to slightly change the generated models and api.

examples/mr_service_example.go
lampajr: no longer used in the tests

internal/core/api.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// type ListRes[T any] struct {
+// 	Items         []*T
+// 	PageSize      *int32
+// 	Size          *int32
+// 	NextPageToken *string
+// }
+
+type ModelRegistryApi interface {
```
tarilabs: In this API, we opted to use `[]*openapi.<entity>, ListResult` as part of the return tuple; alternatively, we thought about this structure. We thought about leaving it as a comment for the evaluation.

dhirajsb: We could also re-use the `*List` types from `internal/model/openapi`. It'll be one less thing to map in `api_model_registry_service_service.go`

test/bdd/features/example.feature
```diff
@@ -0,0 +1,9 @@
+Feature: As a MLOps engineer I would like the Model Registry to store metadata information about models
+  Taken from User Stories
+
+  Scenario: As a MLOps engineer I would like to store Model name
+    Given I have a connection to MR
+    When I store a RegisteredModel with name "Pricing Model" and a child VersionedModel with name "v1" and a child Artifact with uri "s3://12345"
```
dhirajsb: Replace `VersionedModel` with `ModelVersion`. There are also other places in the code where it needs to be changed.

Makefile
```diff
@@ -119,6 +119,10 @@ lint: gen
 test: gen
 	go test ./internal/...
 
+.PHONY: clean-test
```
dhirajsb: The target name `clean-test` is confusing because `clean` is another target in the file. Can we use a better name to be explicit?

dhirajsb (changes requested):
Added comments for re-using `*List` types, make target, and renaming `VersionedModel` in tests. 
Otherwise it looks good. Nice work with testcontainers. :+1:

test/bdd/features/example.feature
tarilabs: Ah, I missed it, sorry. Nice catch! 🏗️

tarilabs: done with b4ec1466043e561c2541ba4402878fbcbde24783

Makefile
lampajr: you are definitely right, it could be quite confusing.

Updated to a more explicit `test-nocache` with 515228e1a02769bbf442c7fd9a8a511b5e28196f.

tarilabs:
> We could also re-use the *List types from internal/model/openapi. It'll be one less thing to map in api_model_registry_service_service.go

Done with c6449f4

dhirajsb (approved):
lgtm

internal/core/core.go
```diff
@@ -0,0 +1,456 @@
...
+		return nil, err
+	}
+
+	return modelVer, nil
+}
+
+func (serv *modelRegistryService) GetModelVersionByParams(name *string, externalId *string) (*openapi.ModelVersion, error) {
+	filterQuery := ""
+	if name != nil {
+		filterQuery = fmt.Sprintf("name = \"%s\"", *name)
```
antisaling, line 260: The client doesn't know that we're prefixing the ModelVersion name with a RegisteredModel id in MLMD

lampajr: That is correct but for now we "assumed" that it is known.
Once https://github.com/opendatahub-io/model-registry/issues/76 got fixed we can use the `registeredModelId` to ignore the previous assumption and let user provide just model version.

lampajr:
As agreed yesterday, going to merge as first service layer iteration.

PR #72, opened 2023-10-16
Setup image build and push to quay

### LINKED ISSUE #15: [feature] Create container image in Quay
Need a container image in Quay to be able to easily deploy the server in a cloud environment. 

A quick and easy image is needed to start in Quay. It should use a simple `Dockerfile` to be able to stand up a model-registry server. 
The container shouldn't need anything else besides the server executable and some default params or config to get it going. To make it easy to get started the server executable should default to a SQLite database. 

@rareddy do we need an org for MR in Quay, or is there an ODH org we should re-use? We'll also need credentials to push our images to Quay. 

None

**Additional context**
This image is a prerequisite for creating a controller to deploy the model-registry in ODH platform.

lampajr:
Fixes https://github.com/opendatahub-io/model-registry/issues/15

## Description
Created two GitHub workflows automating the container image build and push to quay.io registry.
I preferred to keep `main` and `pr` checks separated to avoid adding conditional workflows that could become confusing and hard to maintain.

Therefore I created a separated `build-image` action that takes care of building the image and pushing it to quay if needed. Then both workflows use that action with different inputs.
 
Main differences between the two workflows:
- `main`: push the generated image using with three different tags, `latest`, `main` and `<commit>`
- `pr`: performs a build for the PR branch and skip the whole push step.

## How Has This Been Tested?

I tested the `main` build and push using local [nektos/act](https://github.com/nektos/act) tool and using my own quay.io credentials.

```bash
...
8e2bc7aea200: Layer already exists 
a94ab5ec3980: Layer already exists 
9aaca8eae7c0: Layer already exists 
latest: digest: sha256:83dd6f300b6010975f7dde18823c13a1fbf5005596182efd198c5da65106b31a size: 948
| The push refers to repository [quay.io/***/model-registry]
8e2bc7aea200: Layer already exists 
a94ab5ec3980: Layer already exists 
9aaca8eae7c0: Layer already exists 
main: digest: sha256:83dd6f300b6010975f7dde18823c13a1fbf5005596182efd198c5da65106b31a size: 948
[Main container image build and tag/build-image]   ✅  Success - Main Tag latest
[Main container image build and tag/build-image] ⭐ Run Post Build Image
[Main container image build and tag/build-image] ⭐ Run Post Login to Quay.io
[Main container image build and tag/build-image]   🐳  docker exec cmd=[node /var/run/act/actions/docker-login-action@v2/dist/index.js] user= workdir=
[Main container image build and tag/build-image]   💬  ::debug::Exec.getExecOutput: docker logout quay.io
| [command]/usr/bin/docker logout quay.io
| Removing login credentials for quay.io
[Main container image build and tag/build-image]   ✅  Success - Post Login to Quay.io
[Main container image build and tag/build-image]   ✅  Success - Post Build Image
[Main container image build and tag/build-image] 🏁  Job succeeded
```

Here the pushed tags in my own repository on quay.io https://quay.io/repository/alampare/model-registry?tab=tags

## Merge criteria:


.github/workflows/build-image-pr.yml
```diff
@@ -0,0 +1,34 @@
+name: Test container image build
+
+on:
+  pull_request_target:
```
lampajr, line 4: As per my experience I had to set `pull_request_target` in order to access repository secrets, therefore this workflow is not going to be triggered in this PR as it needs to be merged first for security reasons.

.github/workflows/build-image-pr.yml
```diff
@@ -0,0 +1,34 @@
...
+    branches: [ '*' ]
+
+env:
+  IMAGE_REPO_MR: model-registry
+  QUAY_ORG: opendatahub
+  QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+  QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
```
tarilabs: I agree the "-pr" to not push to Quay, I understand you might need to supply some values for `QUAY_USERNAME` and `QUAY_PASSWORD` but to avoid security-related-questions, can't we either omit them completely here or set them to some dummy values? See other comment about `action.yml`

.github/actions/build-image/action.yml
```diff
@@ -0,0 +1,71 @@
...
+  using: "composite"
+  steps:    
+    - uses: actions/checkout@v3
+      with:
+        path: build
+    - name: Login to Quay.io
+      uses: docker/login-action@v3
+      with:
+        username: ${{ env.QUAY_USERNAME }}
+        password: ${{ env.QUAY_PASSWORD }}
```
tarilabs: to avoid the need of login, shouldn't we guard this with `SKIP_IF_EXISTING == 'true' || inputs.PUSH == 'true'` ?

That way, if the caller GHA does set
```yaml
        with:
          SKIP_IF_EXISTING: false
// ...
          PUSH: false
```

as the `-pr.yml` action below, you don't need to supply user/psw?

tarilabs (commented):
Thank you so much!!

Just some open questions:

.github/actions/build-image/action.yml
lampajr: I definitely agree on this, as long as we don't push anything it would be better to completely skip that step.
Going to apply the fix

lampajr: acutally I think that `inputs.PUSH == 'true'` is enough as guard.

tarilabs: > acutally I think that `inputs.PUSH == 'true'` is enough as guard.

Oh I see, the "Check if Image already exists" uses curl directly, not docker/podman commands; I agree

tarilabs (dismissed):
thank you so much

.github/actions/build-image/action.yml
lampajr: Resolved by 5548703ce333854a319650a016b672b1707cac96. Given that we skip the login I reverted the trigger event to `pull_request`

.github/workflows/build-image-pr.yml
lampajr: Resolved by 5548703ce333854a319650a016b672b1707cac96

tarilabs: was resolved with 5548703

.github/actions/build-image/action.yml
tarilabs: was resolved with 5548703

.github/workflows/build-image-pr.yml
lampajr: As we don't push anything from PR checks we don't even need to login to registry, therefore we do not need to access the secrets.

Reverted to `pull_request` event.

.github/actions/build-image/action.yml
```diff
@@ -0,0 +1,72 @@
...
+    # Case 4: >1 tags were created, but the most recent one was never deleted (same as case 3, but the latest tag does not have "end_ts".
+    - name: Check if Image already exists
+      shell: bash
+      if: inputs.SKIP_IF_EXISTING == 'true'
+      env:
+        IMAGE: quay.io/${{ env.QUAY_ORG }}/${{ inputs.IMAGE_REPO }}:${{ env.TARGET_IMAGE_TAG }}
+      run: |
+        echo ${{ inputs.SKIP_IF_EXISTING }}
+
+        tags=$(curl --request GET 'https://quay.io/api/v1/repository/${{ env.QUAY_ORG }}/${{ inputs.IMAGE_REPO }}/tag/?specificTag=${{ env.TARGET_IMAGE_TAG }}')
```
dhirajsb: @lampajr let's put this logic in a shell script. That'll make it re-usable in a non-github action build like downstream and managed services builds.

dhirajsb (changes requested):

.github/actions/build-image/action.yml
lampajr: you mean the whole step or just the command to get all tags `tags=$(curl --request GET 'https://quay.io/api/v1/repository/${{ env.QUAY_ORG }}/${{ inputs.IMAGE_REPO }}/tag/?specificTag=${{ env.TARGET_IMAGE_TAG }}')` ? I guess the latter right?

dhirajsb: I was thinking we could create something like rhosak [build_deploy.sh](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/build_deploy.sh)

lampajr: I see your point and make sens to have a script doing the same build and push - I am working on this. Moreover I'd like to incorporate some of the sub tasks in the Makefile as well (like `docker build` and `docker push`) similarly to what has been done in the script you shared

.github/workflows/build-image-pr.yml
```diff
@@ -0,0 +1,27 @@
+name: Test container image build
+
+on:
+  pull_request_target:
```
lampajr, line 4: As we are going to run an internal script that can pushes on the registry I would prefer to keep `pull_request_target` to avoid malicious code execution on external PR.

Reference: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

```
This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow
```

.github/actions/build-image/action.yml
lampajr: I was able to extract the build & deploy stuff in a dedicated script, I also put some of the tasks in the Makefile as anticipated.
Given that I could safely remove the action as not needed anymore, for more details check commit de310c181d2e555141102dfc8230558499aba3aa.

.github/workflows/build-image-pr.yml
lampajr: Therefore the PR build image workflow is not going to be executed in this PR, as it need to get merged first.
I tried locally and it worked:
```bash
act --env QUAY_ORG=alampare -W .github/workflows/build-image-pr.yml
```

image

dhirajsb (approved):
/lgtm

PR #77, opened 2023-10-20
Add Python client

antisaling:
Add a preliminary Python client implementation.

## Description
I've based my implementation on #68 and our OpenAPI definition.

## How Has This Been Tested?
There are simple tests included. I still have to add instructions to run, a CI job to run them
automatically, and docstrings.

To test, first make sure you have at least Python 3.10 installed, if you don't:
1. Setup [`pyenv`](https://github.com/pyenv/pyenv)
2. `pyenv install 3.10.13`

Now, install [`nox`](https://github.com/wntrblm/nox), then, to test, follow the instructions on the README.md file.

## TODO:
- [x] Document `nox`, etc on the README
- [x] Build Docs
- [x] Add CI jobs for building docs and running tests
- [ ] Release the first version

## Merge criteria:

- [x] The commits and have meaningful messages
- [ ] The author will squash them [after approval](https://github.com/opendatahub-io/opendatahub-community/blob/main/contributor-cheatsheet.md#:~:text=Usually%20this%20is%20done%20in%20last%20phase%20of%20a%20PR%20revision) or will ask to merge with squash.

I don't know if it makes sense to squash this, we can discuss that.


dhirajsb:
@isinyaaa I think one of the suggestions from @rimolive from DSP team was that we could re-use mlmd base classes and extend our types from them. 
Would that make it possible to inherit common properties and perhaps make it easier to implement the API somehow? 
I am not sure about all the implementation details. @rimolive can you provide more information about what you were thinking?

antisaling:
v2 updates:
- Moved the code into a `src` folder as per guidelines.
- Removed `map` submodule: let's keep things simple with a single interface and helper methods.
- Moved model registry client into submodule
- Created a new interface for prefixable types
- Simplified artifact state mapping (removed bidict dependency)
- Added simple docstrings
- Created a noxfile to simplify running tasks
- Better separation of core functionality from MLMD integration/dependency
- Dealt with TODOs regarding MLMD errors
- Added proper UUID prefixing
- Auto-register `Base{Artifact,Context}` subclasses using `__init_subclass__`
- Add back `model` when fetching `ModelVersion`
- Added pytest coverage plugin
- Improved test coverage (now at 87%) and logic

dhirajsb (dismissed):
lgtm

Just to clarify, is the primary purpose of the store wrapper type for parity with the Golang client's service layer API?

antisaling:
> lgtm
> 
> Just to clarify, is the primary purpose of the store wrapper type for parity with the Golang client's service layer API?

I'm focusing on parity on the base type definitions and on the registry client. The store wrapper should simply provide a useful interface to gRPC focusing on our use case (upserting and querying single protos).

clients/python/src/model_registry/types/base.py
```diff
@@ -0,0 +1,145 @@
...
+        name (str): Customizable name. Has to be unique among instances of the same type.
+        external_id (str, optional): Customizable ID. Has to be unique among instances of the same type.
+        create_time_since_epoch (int): Seconds elapsed since object creation time, measured against epoch.
+        last_update_time_since_epoch (int): Seconds elapsed since object last update time, measured against epoch.
+    """
+    id: str
+    name: str
+    external_id: Optional[str]
+    create_time_since_epoch: int
+    last_update_time_since_epoch: int
```
antisaling: TBH I can't really see the usefulness of providing these fields to the client at the cost of an additional query. The only potential use case I see is some way of sorting by timestamp, but `ListOptions` already provides this functionality when querying in bulk on the server side.

dhirajsb: Yeah, unfortunately gRPC only returns the ids on create calls. Let's go ahead and do the extra call because we will have default values for some _system_ properties beside these timestamps. And, the client library should return those defaults to the user. 
For example, we'll soon need to add a creator/owner username property to show who created which models, etc. in the UI. That property will also be needed for RBAC in the future.

antisaling:
Changes in v3:
- Added sphinx to build docs
- Added basic usage on README
- Added attrs
- Made exceptions a common submodule
- Brought back client as single file module as we don't need much more right now
- Fixed license string
- Added CI to build docs and test
- Added release commit
- Fixed (most) `get_*` calls to automatically populate references to other proto objects.(Artifacts are still pending)
- Fixed ListOptions
- Fixed `ModelRegistry` constructor to take in SSL connection args
- Renamed `*Error` -> `*Exception` following python conventions
- Added description fields to all MLMD wrapper types
- Some more improvements to tests

clients/python/pyproject.toml
```diff
@@ -0,0 +1,44 @@
+[tool.poetry]
```
dtrifiro, line 1: Did you encounter any issues so far by using poetry?

clients/python/src/model_registry/__init__.py
```diff
@@ -0,0 +1,12 @@
+"""Main package for the ODH model registry."""
+
+__version__ = "0.1.0"
```
dtrifiro, line 3: Also related to the above: I find it rather cumbersome to have version numbers hardcoded in code and/or `pyproject.toml`.

To solve this issue, I like building with setuptools and using [`setuptools_scm`](https://github.com/pypa/setuptools_scm/), which makes it easy to retrieve version information from git. 

If you like to stick with poetry, something that **might** work with poetry is this [`poetry-dynamic-versioning`](https://github.com/mtkennerly/poetry-dynamic-versioning/), although I have not used it before. If you want to do some reading, there's quite a few issues in the poetry repo related to this problem.

clients/python/src/model_registry/client.py
```diff
@@ -0,0 +1,408 @@
...
+        Args:
+            options (ListOptions, optional): Options for listing registered models.
+
+        Returns:
+            Sequence[RegisteredModel]: Registered models.
+        """
+        if options is None:
+            mlmd_options = None
+        else:
+            mlmd_options = options.as_mlmd_list_options()
```
dtrifiro:
```suggestion
       mlmd_options = options.as_mlmd_list_options() if options else None
```

clients/python/src/model_registry/client.py
```diff
@@ -0,0 +1,408 @@
...
+    # TODO: does this call make sense?
+    def get_model_artifacts(
+        self,
+        model_version_id: Optional[str] = None,
+        options: Optional[ListOptions] = None,
+    ) -> Sequence[ModelArtifact]:
+        if options is None:
+            mlmd_options = MLMDListOptions()
+        else:
+            mlmd_options = options.as_mlmd_list_options()
```
dtrifiro: Same as above

clients/python/src/model_registry/client.py
```diff
@@ -0,0 +1,408 @@
...
+        """Fetch a model artifact either by external ID or by the ID of its associated model version.
+
+        Args:
+            model_version_id (str, optional): ID of the associated model version.
+            external_id (str, optional): Model artifact external ID.
+
+        Returns:
+            ModelArtifact: Model artifact.
+        """
+        if external_id is not None:
```
dtrifiro: How should this behave for an empty string? 
```suggestion
        if external_id:
```

using this will not call `get_artifact` with an empty id

clients/python/src/model_registry/exceptions.py
```diff
@@ -0,0 +1,32 @@
+"""Exceptions for the model registry.
+"""
+
+
+class StoreException(Exception):
+    """Storage related error."""
+
+    pass
```
dtrifiro: This pass (and all of the following) are not required

clients/python/src/model_registry/store/wrapper.py
```diff
@@ -0,0 +1,314 @@
...
+            except errors.NotFoundError as e:
+                raise TypeNotFoundException(
+                    f"Context type {type_name} does not exist"
+                ) from e
+            except errors.InternalError as e:
+                raise ServerException(
+                    f"Couldn't get context type {type_name} from MLMD store"
+                ) from e
+        else:
+            raise UnsupportedTypeException(f"Unsupported type: {type_name}")
```
dtrifiro:
```suggestion
        if isinstance(mlmd_pt, Artifact):
            name = "Artifact"
            method =  self._mlmd_store.get_artifact_type
        elif isinstance(mlmd_pt, Context):
            name = "Context"
            method = self._mlmd_store.get_context_type 
        else:
            raise UnsupportedTypeException(f"Unsupported type: {type_name}")
 
        try:
            _type =method(type_name)
        except errors.NotFoundError as e:
            raise TypeNotFoundException(
                f"{name} type {type_name} does not exist"
            ) from e
         except errors.InternalError as e:
            raise ServerException(
                f"Couldn't get {name.lower()} type {type_name} from MLMD store"
            ) from e
```

How about something like this to avoid duplicating code?

clients/python/src/model_registry/store/wrapper.py
```diff
@@ -0,0 +1,314 @@
...
+            ids = self._mlmd_store.put_artifacts([artifact])
+        except errors.AlreadyExistsError as e:
+            raise DuplicateException(f"Artifact {artifact.name} already exists") from e
+        except errors.InvalidArgumentError as e:
+            raise StoreException("Artifact has invalid properties") from e
+        except errors.NotFoundError as e:
+            raise TypeNotFoundException(
+                f"Artifact type {artifact.type} does not exist"
+            ) from e
+        return ids[0]
```
dtrifiro:
```suggestion
        try:
            return self._mlmd_store.put_artifacts([artifact])[0]
        except errors.AlreadyExistsError as e:
            raise DuplicateException(f"Artifact {artifact.name} already exists") from e
        except errors.InvalidArgumentError as e:
            raise StoreException("Artifact has invalid properties") from e
        except errors.NotFoundError as e:
            raise TypeNotFoundException(
                f"Artifact type {artifact.type} does not exist"
            ) from e
```

clients/python/src/model_registry/types/base.py
```diff
@@ -0,0 +1,204 @@
...
+            if isinstance(value, bool):
+                mlmd_props[key].bool_value = value
+            elif isinstance(value, int):
+                mlmd_props[key].int_value = value
+            elif isinstance(value, float):
+                mlmd_props[key].double_value = value
+            elif isinstance(value, str):
+                mlmd_props[key].string_value = value
+            else:
+                raise Exception(f"Unsupported type: {type(value)}")
```
dtrifiro, line 153: Seems like a good place to use structural pattern matching: https://docs.python.org/3/whatsnew/3.10.html?highlight=match#pep-634-structural-pattern-matching

clients/python/src/model_registry/types/contexts.py
```diff
@@ -0,0 +1,107 @@
...
+
+from .artifacts import BaseArtifact, ModelArtifact
+from .base import Prefixable, ProtoBase
+
+
+@define(slots=False, init=False)
+class BaseContext(ProtoBase, ABC):
+    """Abstract base class for all contexts."""
+
+    pass
```
dtrifiro: useless pass statement

clients/python/tests/store/test_wrapper.py
```diff
@@ -0,0 +1,131 @@
...
+        Context: Created context.
+    """
+    ctx_type = ContextType()
+    ctx_type.name = name
+
+    ctx = Context()
+    ctx.name = name
+    ctx.type_id = store_wrapper._mlmd_store.put_context_type(ctx_type)
+
+    return ctx
```
dtrifiro: These could be pytest fixtures

dtrifiro (approved):
Looks great!

clients/python/pyproject.toml
antisaling: I had some issues with Python versions, and it's also not as clear how to get proper integration with an editor (I've been using `VenvSelector` on vim with some hiccups on my Mac).

clients/python/src/model_registry/__init__.py
antisaling: I'd had a look at that and thought it was unnecessary for now, as we are still iterating on an alpha version. This plugin you sent seems to rely on git tags, which we can't use, unless we share tags for the go MR as well as the python client.

clients/python/pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+name = "model-registry"
+version = "0.1.0"
+description = "Client for Red hat OpenDataHub Model Registry"
```
tarilabs:
```suggestion
description = "Client for OpenDataHub Model Registry"
```

Very minor, but I think being the community version, they would prefer not to prefix it, to inadvertently give wrong impression?
If you want to keep it, it's fine too, but capital H per brand guidelines :) `Red Hat`

tarilabs (dismissed):
Imho you can consider Dhiraj and Daniele's comments in further PR, if you prefer to merge this now, and could be further improved directly on main for their suggestions and further dev work.

For the records for QE and others;
I've tested this both in DevContainer (I personally find that convenient for IDE support):

Screenshot 2023-11-06 at 14 59 20

and Rosetta emulation:
```
$env /usr/bin/arch -x86_64 /bin/zsh --login
# then
poetry install
poetry build
poetry run pytest
```

clients/python/src/model_registry/store/wrapper.py
antisaling: Thx for the suggestion, I didn't bother too much with that as we're only dealing with those two. There must also be better ways of mapping/unmapping.

clients/python/tests/store/test_wrapper.py
antisaling: I couldn't find a simple way to parameterize names for those, but now I see this is not necessary.

clients/python/src/model_registry/types/base.py
antisaling: I had tried this before only to get `Irrefutable pattern is allowed only for the last case statement`, but now I see [I should be matching against an instance](https://stackoverflow.com/questions/72295812/python-match-case-by-type-of-value) and not the type itself.

clients/python/pyproject.toml
antisaling: Oh MB :) thx for the suggestion.

antisaling:
Changes in v4:
- Applied suggestions
- Used structural pattern matching where possible

clients/python/src/model_registry/types/base.py
antisaling: Unfortunately it looks like this isn't supported in 3.9... I wouldn't mind dropping compat for this version if we were able to use 3.11, but the MLMD py lib only goes so far as 3.10. Not sure if it's OK to only provide compat for a single version...

antisaling:
Changes in v5:
- Dropped structural pattern matching to keep compatibility with 3.9.

tarilabs (approved):

Issue #93, opened 2023-10-26
Skip PR checks when not needed


lampajr:
If we are changing files for which we do not have to perform any kind of check, e.g, if we are changing ISSUE_TEMPLATE or just doc files, we should skip GitHub workflow trigger.

tarilabs:
+1, as @isinyaaa previously suggested we could ignore GHA running when modification happens in specific directories, or specific file extensions (eg: `.md`)

PR #94, opened 2023-10-26
Create dependabot.yml

dhirajsb:
## Description

Enable dependabot version updates.

## How Has This Been Tested?

NA

## Merge criteria:

~~- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).~~
~~- [ ] The developer has manually tested the changes and verified that the changes work~~

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,10 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
   - package-ecosystem: "docker"
     directory: "/"
     schedule:


dhirajsb:
@lampajr we should add some path filters to avoid building image for PRs like this one that doesn't affect any image sources.

tarilabs:
> we should add some path filters to avoid building image for PRs

i believe this covers: https://github.com/opendatahub-io/model-registry/issues/93

lampajr:
> @lampajr we should add some path filters to avoid building image for PRs like this one that doesn't affect any image sources.

I definitely agree on this, I opened an issue few mins ago https://github.com/opendatahub-io/model-registry/issues/93

tarilabs (commented):
question: do you want to consider as well adding Python with this PR, or tracking it in another issue?

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
lampajr, line 11: do we want all version updates or just security ones for now?
if we go for the latter, it would be enough to add `open-pull-requests-limit: 0` as suggested https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates#overriding-the-default-behavior-with-a-configuration-file

if we go for the former, it looks great to me!

tarilabs: you can also consider adding:
```yaml
    allow:
      - dependency-type: "production"
```

to account only for non-dev-time dependencies, based on experiences from other projects, up to considerations

dhirajsb: Let's update all dependencies where we can, so we avoid waiting to run into issues because of an older dependency. There are also a bunch of vulnerabilities (most of which seem to be c libraries from sqlite dependency iiuc) being reported in quay repo that we'll need to address. 
Production is only supported for python in our packages anyway, so I'm leaning towards not using that right now.

dhirajsb:
Added pip package for python dependencies.

lampajr (approved):

rareddy:
Nice! this would catch any existing CVE issues.

tarilabs:
> Nice! this would catch any existing CVE issues

@rareddy fwiw we had them already:

- https://github.com/opendatahub-io/model-registry/pull/86
- https://github.com/opendatahub-io/model-registry/pull/69

this PR, in my understanding, will override Org level settings for specific MR settings (works for me I assume we wanted fine tuned controls).

PR #112, opened 2023-10-30
workflows: fix branch restriction on pull requests

antisaling:
As we actually want to run those workflows for any branches, we should either specify "**" to include names with slashes (/) or simply omit the restriction.


lampajr (approved):
lgtm, thx a lot for spotting it!

PR #113, opened 2023-10-30
Improve core layer optional params management

### LINKED ISSUE #105: Simplify check in Core layer impl
followup to https://github.com/opendatahub-io/model-registry/pull/85#discussion_r1372857213

### LINKED ISSUE #110: Make use of glog in core layer
As suggested [here](https://github.com/opendatahub-io/model-registry/pull/85#discussion_r1374915916) make use of [`glog`](https://github.com/golang/glog).

Substitute standard `log` module with `glog` one.

n/a

**Additional context**
n/a

lampajr:
Fixes https://github.com/opendatahub-io/model-registry/issues/105 
Fixes https://github.com/opendatahub-io/model-registry/issues/110
Furthermore it provides some optional paramas management 

## Description

Additional improvements:
- For `Update` operations, user does not have to provide the `parentResourceId` - the implementation is going to retrieve it from the original parent association.
  - Created two internal utlities: `getModelVersionByArtifactId` and `getRegisteredModelByVersionId`
- Manage all possible failures that can happen for `create/update`, based on provided input params.

## How Has This Been Tested?

Added additional test cases in `core_test` that cover the new edge cases, mostly considering possible failures.
```bash
make test-cover
```
## Merge criteria:

### DIFF
--- a/internal/core/core.go
+++ b/internal/core/core.go
@@ -79,10 +79,26 @@ func NewModelRegistryService(cc grpc.ClientConnInterface) (ModelRegistryApi, err
 // REGISTERED MODELS
 
 func (serv *modelRegistryService) UpsertRegisteredModel(registeredModel *openapi.RegisteredModel) (*openapi.RegisteredModel, error) {
+	var err error
+	var existing *openapi.RegisteredModel
+
 	if registeredModel.Id == nil {
-		log.Printf("Creating registered model for %s", *registeredModel.Name)
+		log.Printf("Creating new registered model")
 	} else {
-		log.Printf("Updating registered model %s for %s", *registeredModel.Id, *registeredModel.Name)
+		log.Printf("Updating registered model %s", *registeredModel.Id)
+		existing, err = serv.GetRegisteredModelById(*registeredModel.Id)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// if already existing assure the name is the same
+	if existing != nil {
+		if registeredModel.Name == nil {
+			// user did not provide it
+			// need to set it to avoid mlmd error "context name should not be empty"
+			registeredModel.Name = existing.Name
+		}
 	}
 
 	modelCtx, err := serv.mapper.MapFromRegisteredModel(registeredModel)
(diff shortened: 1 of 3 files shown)


internal/core/core.go
```diff
@@ -79,10 +80,26 @@ func NewModelRegistryService(cc grpc.ClientConnInterface) (ModelRegistryApi, err
...
+		}
+	}
+
+	// if already existing assure the name is the same
+	if existing != nil {
+		if registeredModel.Name == nil {
+			// user did not provide it
+			// need to set it to avoid mlmd error "context name should not be empty"
+			registeredModel.Name = existing.Name
+		}
```
tarilabs:
```suggestion
	if existing != nil && registeredModel.Name == nil {
		// user did not provide it
		// need to set it to avoid mlmd error "context name should not be empty"
		registeredModel.Name = existing.Name
	}
```

how about? 🤔

internal/core/core.go
```diff
@@ -200,26 +254,48 @@ func (serv *modelRegistryService) GetRegisteredModels(listOptions ListOptions) (
...
-	modelCtx, err := serv.mapper.MapFromModelVersion(modelVersion, *registeredModelIdCtxID, registeredModelName)
+
+	// if already existing assure the name is the same
+	if existing != nil {
+		if modelVersion.Name == nil {
+			// user did not provide it
+			// need to set it to avoid mlmd error "context name should not be empty"
+			modelVersion.Name = existing.Name
+		}
+	}
```
tarilabs: Same here, how about

```suggestion
	if existing != nil && registeredModel.Name == nil {
		// user did not provide it
		// need to set it to avoid mlmd error "context name should not be empty"
		registeredModel.Name = existing.Name
	}
```

tarilabs (dismissed):
Thank you!
Minor comments, up to you.

Later when wiring the REST, we could think of specializing error with types beyond the string-msg, so to return proper HTTP xxx error codes; wdyt?

internal/core/core.go
lampajr: good catch!! forgot to simplify it :man_facepalming:

lampajr:
> Later when wiring the REST, we could think of specializing error with types beyond the string-msg, so to return proper HTTP xxx error codes; wdyt?

Definitely agree on that, we can create our own `Error` instances wrapping additional meaningful fields

internal/core/core.go
lampajr: hehe it should have been `modelVersion` and not `registeredModel`, fixed with 5c10d074a9537a74fc051fc5cee21ef098fcf4da

lampajr:
There was an issue in the applied suggestion, just a typo. 
Fixed with 5c10d074a9537a74fc051fc5cee21ef098fcf4da

tarilabs (approved):

PR #139, opened 2023-11-10
Create BuildListOption for REST query params

### LINKED ISSUE #138: Missing query params for pagination should be ignored
**Describe the bug**
Query params for pagingation like `pageSize` are not always required. But if they are missing it causes an error in the server. 

**To Reproduce**
1. Use curl to call REST endpoint without `pageSize`
2. Server sends an error response with `{"code":"","message":"strconv.Atoi: parsing \"\": invalid syntax"}`

**Expected behavior**
If not provided, use default some default values for pagination parameters, .e.g default `pageSize` of 20. 

**Additional context**
Consult default values used by mlmd in proto/code.

tarilabs:
Resolves #138 
as discussed with @lampajr 

## Description
Ignores unsupplied-but-golang-empty query params, and relies on defaults by gRPC; the defaults could now be overridden for alternative default-values in a single place if needed later.

## How Has This Been Tested?
After Build and a MLMD available on docker:
```
robot test/robot/UserStory.robot
curl -X 'GET' 'http://localhost:8080/api/model_registry/v1alpha1/registered_models?pageSize=47' -H 'accept: application/json'
curl -X 'GET' 'http://localhost:8080/api/model_registry/v1alpha1/registered_models' -H 'accept: application/json'
```

## Merge criteria:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,7 @@ COPY cmd/ cmd/
 COPY api/ api/
 COPY config/ config/
 COPY internal/ internal/
+COPY pkg/ pkg/
 
 # Build
 USER root
--- a/internal/apiutils/api_utils.go
+++ b/internal/apiutils/api_utils.go
@@ -46,12 +46,12 @@ func ZeroIfNil[T any](input *T) T {
 	return *new(T)
 }
 
-func BuildListOption(pageSize string, orderBy model.OrderByField, sortOrder model.SortOrder, nextPageToken string) (*api.ListOptions, error) {
+func BuildListOption(pageSize string, orderBy model.OrderByField, sortOrder model.SortOrder, nextPageToken string) (api.ListOptions, error) {
 	var pageSizeInt32 *int32
 	if pageSize != "" {
 		conv, err := converter.StringToInt32(pageSize)
 		if err != nil {
-			return nil, err
+			return api.ListOptions{}, err
 		}
 		pageSizeInt32 = &conv
 	}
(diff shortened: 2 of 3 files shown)


internal/core/api_utils.go
```diff
@@ -41,3 +43,32 @@ func BuildListOperationOptions(listOptions ListOptions) (*proto.ListOperationOpt
 	}
 	return &result, nil
 }
+
+func BuildListOption(pageSize string, orderBy model.OrderByField, sortOrder model.SortOrder, nextPageToken string) (*ListOptions, error) {
```
lampajr: Why not returning `ListOptions` rather than a pointer to it? in that way we can avoid dereferencing it like here `s.coreApi.GetInferenceServices(*listOpts, &servingenvironmentId)` as in all usages we are passing the value.

lampajr (dismissed):
lgmt, left just one minor comment

internal/core/api_utils.go
tarilabs: Originally I did it for consistency with other builders, but I concur I prefer the value semantic too.
Done 👍

lampajr (dismissed):

lampajr (approved):
Rebased from `main` and added commit a55f7bea5b7dd7ffa29114c3405b5ca2e782f2a1 to fix the image container build issue.

Note that the check here is still failing because for security reason we are using `pull_request_target` event, this way changes applied on this PR are not actually tested in the container build --> raised issue https://github.com/opendatahub-io/model-registry/issues/145 to keep track of this.

The change lgtm!

PR #146, opened 2023-11-14
python check: do not ignore the workflow itself

lampajr:
## Description

Do not ignore the python check workflow, we should test even when changes are applied there.
An example is https://github.com/opendatahub-io/model-registry/pull/143 where we are upgrading an action used by python check and no check is run to test it.

## How Has This Been Tested?

Now in this PR, python tests are correctly triggered.

## Merge criteria:


antisaling (approved):
lgtm, thanks for the patch Andrea :)

PR #153, opened 2023-11-15
Update build-image-pr.yml: switch to pull_request event

### LINKED ISSUE #145: [ci] PR container build check not testing PR changes
For security reason in `build-image-pr.yml` workflow we are using `pull_request_target` event [1], unfortunately in this way it seems that pull request changes (e.g., affecting the Dockerfile) are not really tested because the the check keeps using the content on `main` branch.

For this reason PR https://github.com/opendatahub-io/model-registry/pull/140 introduced an issue in the Dockerfile even if the test container check was green. That is because the test was still performed against `main` branch.

**Solution**:
We should probably switch to `pull_request` event in order to really check PR content

**Alternatives**:
TBD

[1] https://github.com/opendatahub-io/model-registry/blob/main/.github/workflows/build-image-pr.yml#L4

lampajr:
Fixes https://github.com/opendatahub-io/model-registry/issues/145

## Description
Switch to the `pull_request` event in order to be sure the actual PR change is getting tested.

## How Has This Been Tested?

This PR will trigger the new workflow automatically

## Merge criteria:

### DIFF
--- a/.github/workflows/build-image-pr.yml
+++ b/.github/workflows/build-image-pr.yml
@@ -1,7 +1,7 @@
 name: Test container image build
 
 on:
-  pull_request_target:
+  pull_request:
     paths-ignore:
       - 'LICENSE*'
       - '**.gitignore'


lampajr:
@tarilabs as per our discussion this should be enough to fix the issue, furthermore from a security point of view secrets are not passed to workflow triggered by PRs from fork.

tarilabs (approved):
> @tarilabs as per our discussion this should be enough to fix the issue, furthermore from a security point of view secrets are not passed to workflow triggered by PRs from fork.

indeed, based on documentation:

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository

ref: https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#:~:text=With%20the%20exception%20of%20GITHUB_TOKEN%2C%20secrets%20are%20not%20passed%20to%20the%20runner%20when%20a%20workflow%20is%20triggered%20from%20a%20forked%20repository

and on secrets page for GHA:

> They are not passed to workflows that are triggered by a pull request from a fork.

ref: https://github.com/opendatahub-io/model-registry/settings/secrets/actions#:~:text=They%20are%20not%20passed%20to%20workflows%20that%20are%20triggered%20by%20a%20pull%20request%20from%20a%20fork.

So this change looks good to me

Issue #131, opened 2023-11-07
Move `runtime` from Artifact to InferenceService


tarilabs:
As noted by during Scrum 2023-10-07, there is a need to move the `runtime` from currently `Artifact` entity:

https://github.com/opendatahub-io/model-registry/blob/958630a33cac0c5686eb16df5731ab3abcd270e9/api/openapi/model-registry.yaml#L1321-L1334

to the `InferenceService` entity.

dhirajsb:
lgtm

dhirajsb:
@isinyaaa is there a complimentary commit to add the property to `InferenceService`?

antisaling:
@dhirajsb I can fix that too, but this commit belongs to PR #132 .

dhirajsb:
Ok, just making sure we don't forget about moving it to the right place.

PR #154, opened 2023-11-15
Fix core library issue

lampajr:
Trying to showcase the core library usage I identified some issues, this PR aims to fix them.

## Description
Summary:
- Manage error for `*ByParams` operation in a uniform way avoiding `nil` dereference.
- Fix `runtime` movement under `InferenceService` obj related to https://github.com/opendatahub-io/model-registry/issues/131 where it was wrongly created on `ServeModel` + added test to ensure correct mapping.

## How Has This Been Tested?

```bash
make test
```

## Merge criteria:

### DIFF
--- a/pkg/core/core.go
+++ b/pkg/core/core.go
@@ -285,8 +285,12 @@ func (serv *modelRegistryService) GetRegisteredModelByParams(name *string, exter
 		return nil, err
 	}
 
-	if len(getByParamsResp.Contexts) != 1 {
-		return nil, fmt.Errorf("multiple registered models found for name=%v, externalId=%v", *name, *externalId)
+	if len(getByParamsResp.Contexts) > 1 {
+		return nil, fmt.Errorf("multiple registered models found for name=%v, externalId=%v", apiutils.ZeroIfNil(name), apiutils.ZeroIfNil(externalId))
+	}
+
+	if len(getByParamsResp.Contexts) == 0 {
+		return nil, fmt.Errorf("no registered models found for name=%v, externalId=%v", apiutils.ZeroIfNil(name), apiutils.ZeroIfNil(externalId))
 	}
 
 	regModel, err := serv.mapper.MapToRegisteredModel(getByParamsResp.Contexts[0])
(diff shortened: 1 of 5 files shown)


pkg/core/core.go
```diff
@@ -99,7 +100,6 @@ func NewModelRegistryService(cc grpc.ClientConnInterface) (api.ModelRegistryApi,
 			Properties: map[string]proto.PropertyType{
 				"description":      proto.PropertyType_STRING,
 				"model_version_id": proto.PropertyType_INT,
-				"runtime":          proto.PropertyType_STRING,
```
tarilabs, line 102: well spotted thanks :)

pkg/core/core.go
```diff
@@ -1052,8 +1064,12 @@ func (serv *modelRegistryService) GetInferenceServiceByParams(name *string, pare
...
 	}
 
-	if len(getByParamsResp.Contexts) != 1 {
-		return nil, fmt.Errorf("multiple InferenceServices found for name=%v, parentResourceId=%v, externalId=%v", apiutils.ZeroIfNil(name), apiutils.ZeroIfNil(parentResourceId), apiutils.ZeroIfNil(externalId))
+	if len(getByParamsResp.Contexts) > 1 {
+		return nil, fmt.Errorf("multiple inference services found for versionName=%v, parentResourceId=%v, externalId=%v", apiutils.ZeroIfNil(name), apiutils.ZeroIfNil(parentResourceId), apiutils.ZeroIfNil(externalId))
+	}
+
+	if len(getByParamsResp.Contexts) == 0 {
+		return nil, fmt.Errorf("no inference services found for versionName=%v, parentResourceId=%v, externalId=%v", apiutils.ZeroIfNil(name), apiutils.ZeroIfNil(parentResourceId), apiutils.ZeroIfNil(externalId))
```
tarilabs, line 1072: I would suggest in a later PR:
- check if we can get away without `ZeroIfNil` but dereference the pointer (I did not found a way at the time, since this snippet prints the address for instance:
```
	x := 123
	y := &x
	fmt.Printf("%v", y)
```
but we can keep the ZeroIfNil, this is just a minor check worth of doing
- Check if we can refactor some of the methods, similarly to how it was done leveraging `func mapTo[S getTypeIder, T any](s S, id int64, convFn func(S) (*T, error)) (*T, error) `

tarilabs (approved):

pkg/core/core.go
lampajr: I agree, worth to check if we can make some easy improvement on core layer side, created https://github.com/opendatahub-io/model-registry/issues/156 to keep track of this.

PR #160, opened 2023-11-16
Use canAddFields option in mlmd type setup

### LINKED ISSUE #159: Use `can_add_fields` option when creating/updating mlmd types
**Describe the bug**
If a DB exists with an older version of an mlmd type being used in the registry and a new property is being added, it throws the following error:

```
W1116 16:35:16.810298    10 metadata_store_service_impl.cc:219] PutContextType failed: Type already exists with different properties: can_add_fields is false while the given type has more properties: stored_type: id: 14
name: "odh.InferenceService"
properties {
  key: "description"
  value: STRING
}
properties {
  key: "model_version_id"
  value: INT
}
properties {
  key: "registered_model_id"
  value: INT
}
properties {
  key: "serving_environment_id"
  value: INT
}
; other_type: name: "odh.InferenceService"
properties {
  key: "description"
  value: STRING
}
properties {
  key: "model_version_id"
  value: INT
}
properties {
  key: "registered_model_id"
  value: INT
}
properties {
  key: "runtime"
  value: STRING
}
properties {
  key: "serving_environment_id"
  value: INT
}
```

**To Reproduce**
1. Use the operator to deploy the registry in a PG db in a pvc.
2. Update a type to add a new property
3. This should throw the above error

**Expected behavior**
Registry should automatically update existing mlmd types. 

**Additional context**
Check other options supported in the Put* calls for any other related use cases.

lampajr:
Fixes https://github.com/opendatahub-io/model-registry/issues/159

## Description

Make use of `CanAddFields` property while creating Model Registry types, this way if we are going to add new properties on existing types the proxy won't fail.

Also added some tests ensuring the behavior is the expected one.

> NOTE: while testing it I found out that some errors were not properly managed in the proxy, calling `Fatalf` which calls exit and then return the error. Managed to directly return the error, see commit 3377c412fad4894fac19d9d545b100f20aa2a2f4.

## How Has This Been Tested?

```bash
make test
```

## Merge criteria:

### DIFF
--- a/pkg/core/core_test.go
+++ b/pkg/core/core_test.go
@@ -323,7 +323,7 @@ func TestModelRegistryStartupWithExistingEmptyTypes(t *testing.T) {
 	modelArtifactResp, _ = client.GetArtifactType(ctx, &proto.GetArtifactTypeRequest{
 		TypeName: modelArtifactTypeName,
 	})
-	assertion.NotNilf(modelArtifactResp.ArtifactType, "model version type %s should exists", *modelArtifactTypeName)
+	assertion.NotNilf(modelArtifactResp.ArtifactType, "model artifact type %s should exists", *modelArtifactTypeName)
 	assertion.Equal(*modelArtifactTypeName, *modelArtifactResp.ArtifactType.Name)
 	assertion.Equal(6, len(modelArtifactResp.ArtifactType.Properties))
 
--- a/pkg/core/core_test.go
+++ b/pkg/core/core_test.go
@@ -217,11 +217,143 @@ func registerInferenceService(assertion *assert.Assertions, service api.ModelReg
 	return *createdEntity.Id
 }
 
+func TestModelRegistryStartupWithExistingEmptyTypes(t *testing.T) {
+	ctx := context.Background()
+	assertion, conn, client, teardown := setup(t)
+	defer teardown(t)
+
+	// create all types without props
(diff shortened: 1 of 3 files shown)


tarilabs (approved):

pkg/core/core_test.go
```diff
@@ -217,11 +217,143 @@ func registerInferenceService(assertion *assert.Assertions, service api.ModelReg
...
+		TypeName: modelVersionTypeName,
+	})
+	assertion.NotNilf(modelVersionResp.ContextType, "model version type %s should exists", *modelVersionTypeName)
+	assertion.Equal(*modelVersionTypeName, *modelVersionResp.ContextType.Name)
+	assertion.Equal(4, len(modelVersionResp.ContextType.Properties))
+
+	modelArtifactResp, _ = client.GetArtifactType(ctx, &proto.GetArtifactTypeRequest{
+		TypeName: modelArtifactTypeName,
+	})
+	assertion.NotNilf(modelArtifactResp.ArtifactType, "model version type %s should exists", *modelArtifactTypeName)
```
tarilabs:
```suggestion
	assertion.NotNilf(modelArtifactResp.ArtifactType, "model artifact type %s should exists", *modelArtifactTypeName)
```

tarilabs (dismissed):

tarilabs (approved):

PR #171, opened 2023-11-20
gha: set up go code coverage using codecov.io

### LINKED ISSUE #170: Setup code coverage for Go implementation
Followup the [code coverage setup for python](https://github.com/opendatahub-io/model-registry/pull/163) and setup the same for the Go implementation. 

Take inspiration from this codecov example: https://github.com/codecov/example-go

n/a

**Additional context**
n/a

dtrifiro:
Coses #170

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,4 +51,11 @@ jobs:
               exit 1
           fi
       - name: Unit tests
-        run: make test
+        run: make test-cover
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3.1.4
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+        with:
+          files: coverage.txt
+          fail_ci_if_error: true
--- a/.github/workflows/python-client.yml
+++ b/.github/workflows/python-client.yml
@@ -5,9 +5,6 @@ on:
     branches:
       - "main"
   pull_request:
-    paths:
-      - "clients/python/**"
-      - ".github/workflows/python-client.yml"
 
 jobs:
   tests:
--- a/Makefile
+++ b/Makefile
@@ -166,7 +166,7 @@ test-nocache: gen
 
 .PHONY: test-cover
 test-cover: gen
-	go test ./internal/... ./pkg/... -cover -count=1
+	go test ./internal/... ./pkg/... -cover -count=1 -race -coverprofile=coverage.txt -covermode=atomic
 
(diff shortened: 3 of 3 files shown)


lampajr:
Hey @dtrifiro , first of all thank you very much!!
I have just one question/doubt, right now for PRs that do not touch the python part we are not executing any python test. 

IIUC the code coverage results are compared with the base branch, which actually includes tests for bot Go and Python, then the codecov report could be misleading as in the PR we have the coverage just fo the Go part in this case, am I right?

If my understanding is right, maybe we should run all source code tests (for Go and Python) for every PR, even if no changes are present for one of the two parts. wdyt @dtrifiro @isinyaaa ?

dtrifiro:
This is the first time I try something like this (two languages in the same repo).

I guess the right way to do this is the following: in each workflow we should generate the code coverage results and upload them as pipeline artifacts. Then, when the workflows are finished, we can run another job/workflow which uploads both coverage reports.

At first I guess we can always run both test suites for all PRs, then an improvement could be [caching](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#using-the-cache-action) the code coverage results and restoring them in the step that uploads the code coverage results to avoid re-running the two test suites at all times.

wdyt?

lampajr:
>in each workflow we should generate the code coverage results and upload them as pipeline artifacts. Then, when the workflows are finished, we can run another job/workflow which uploads both coverage reports.

based on the codecov documention it should not be an issue if we upload multiple reports from different workflow, they should merge the reports on their side. See https://docs.codecov.com/docs/merging-reports:
```
Codecov does not override report data for multiple uploads. We always merge the data. Simply upload all three reports at once, or separately.
```

>At first I guess we can always run both test suites for all PRs, then an improvement could be [caching](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#using-the-cache-action) the code coverage results and restoring them in the step that uploads the code coverage results to avoid re-running the two test suites at all times.

Definitely agree on this, we could start running both and then think about optimizing it!

dtrifiro:
I enabled python tests for pull requests even if they do not modify the code. Right now it takes less than a minute to run the full workflow, so I think it's not a large overhead.

lampajr (approved):
lgtm!

Thanks again @dtrifiro

tarilabs (approved):
thank you!

antisaling (approved):
LGTM, thanks a lot @dtrifiro  :)

PR #168, opened 2023-11-18
Fix strict symbol validation in MR Python client

tarilabs:
Resolves #167 

## Description
Currently 2 types of error related to symbol resolution are experienced when using the MR python client on the ODH Jupyter.
Btw @dtrifiro can you kindly help us identify why these are not an issue when running tests locally? 🤔 I'm wondering if it's some opt-in option in strictness for Python interpreter?

The `cannot import name 'ListOptions' ...` error:

Screenshot 2023-11-18 at 21 25 58

The `name 'StoreError' is not defined ...` error:

Screenshot 2023-11-18 at 21 43 41

With these code changes, those errors are no longer experienced when using the MR python client on ODH Jupyter.

## How Has This Been Tested?
Reinstalling the MR python client as a wheel in the Jupyter resolves the documented errors above.

## Merge criteria:

- [x] The commits and have meaningful messages; the author will squash them [after approval](https://github.com/opendatahub-io/opendatahub-community/blob/main/contributor-cheatsheet.md#:~:text=Usually%20this%20is%20done%20in%20last%20phase%20of%20a%20PR%20revision) or will ask to merge with squash.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #180, opened 2023-11-21
python client: improve ci

dtrifiro:
- improve CI by adding a linting step with ruff and mypy. The latter allows for failure, but adds an error message to the workflow status page
- add target-version to ruff config

ideally running ruff linting as part of CI should avoid hitting issues like the ones encountered in #168


dtrifiro:
Merge after #168  and #179

antisaling (approved):

PR #181, opened 2023-11-21
py: add ruff rules

antisaling:
As suggested by @dtrifiro, let's add some Ruff rules to the python client config.

## Description
Add relevant rules to keep python code in check. See https://github.com/isinyaaa/model-registry/blob/878aa46755cc3a0852f792d972b10be0dadc753b/clients/python/pyproject.toml for more details on the rules enabled.

Most new warnings have been fixed as well.

## Merge criteria:


clients/python/pyproject.toml
```diff
@@ -39,7 +39,34 @@ line-length = 119
...
+    "I",   # isort
+    "PT",  # pytest
+    "Q",   # flake8-quotes
+    "RET", # flake8-return
+    "SIM", # flake8-simplify
+    "UP",  # pyupgrade
+]
+ignore = [
+    "E501", # line too long
+    "S101", # use of assert detected
```
dtrifiro, line 67: Not sure if you use this for non-test code, but you can add a per-file-ignore like so:

```toml
[tool.ruff.per-file-ignores]
"__init__.py" = ["F401"]
"tests/*" = ["S101"]
"tests/conftest.py" = ["F403"]
```

clients/python/tests/__init__.py
```diff
@@ -0,0 +1,16 @@
+# ruff: noqa: D101
```
dtrifiro: this could also be added to the per-files-ignore list, as well as the rest, but this is just nitpicking

clients/python/pyproject.toml
dtrifiro: Not sure if you want to add an exception for `D102 Missing docstring in public method` somewhere. CI is failing due to that

clients/python/tests/__init__.py
antisaling: I'll check that option.

clients/python/pyproject.toml
antisaling: Thanks for the suggestion on `per-file-ignores`, see if that makes sense below.

> Not sure if you want to add an exception for D102 Missing docstring in public method somewhere. CI is failing due to that

Resolved using `typing_extensions` to add `@override`.

tarilabs (dismissed):

antisaling:
Rebased to add two commits that I had planned for later. I squashed the current [2/6] with the previous [2/5]. I believe this makes some annotations easier and also simplifies the mapping part quite nicely. I still haven't dealt with all of mypy's complaints though, and I see it isn't running on the CI.

tarilabs (approved):
reconfirming my previous approval [here](https://github.com/opendatahub-io/model-registry/pull/181#event-11050421696) 👍

dtrifiro (approved):

lampajr (approved):
lgtm

PR #182, opened 2023-11-21
py: Use nightly MLMD

antisaling:
Use MLMD nightly so we can extend support to Python 3.11 and also support environments with `attrs>=21.4.0` e.g. ODH notebook images.

Depends on: #181 

## Description
Use https://pypi-nightly.tensorflow.org/simple as the new source for the MLMD package, taking the nightly version from there.

## How Has This Been Tested?
This has been tested on ODH notebook environments, as well as using `nox -s tests`.

## Merge criteria:


dtrifiro (approved):

clients/python/pyproject.toml
```diff
@@ -7,9 +7,15 @@ license = "Apache-2.0"
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = ">= 3.9, < 3.11"
-attrs = "^21.0"
-ml-metadata = "^1.14.0"
+python = ">= 3.9, < 3.12"
+attrs = "^23.1"
+typing-extensions = "^4.8"
+ml-metadata = {version = "^1.15.0.dev20231121", source = "tfx-nightly"}
```
tarilabs, line 13: I'm not against this bump, but I'm conscious of the fact we are in "bootstrap mode" and we didn't have yet a productized version, while at the same time according to this comment: `https://github.com/google/ml-metadata/issues/185#issuecomment-1809561739:~:text=currently%20in%20a%20maintenance` the MLMD project seems in maintenance mode.
So, it's realistic to imagine a 1.15.0 (final) version of MLMD might not be there by our release date.

With this premise, I see the following options.

Option0. We keep this nightly pointer.
If MLMD 1.15 is not released by the time, we will need to provide clear instruction how to install this specific nightly dependency with pip

Option1.
As above, but if 1.15 will not be existing, we will instead "release a custom version" [similarly to this](https://github.com/tarilabs/ml-metadata-remote/tree/v1.14.0/ml_metadata-1.14.0), which btw does not even reference attrs (so a non-attrs dependent, but might be best to double check this with some python SME).

Option2. We do not depend on nightly pointer.
As the tech demo shown, it's possible to install MLMD in jupyter notebook served on the images; we might investiate further without depending on nightly.

Hope this is helpful feedback; wdyt?

Issue #173, opened 2023-11-20
[model-controller] Implement controller for MR and serving interactions


lampajr:
**Is your feature request related to a problem? Please describe.**
Followup of https://github.com/opendatahub-io/model-registry/issues/104: implement the described flow diagram.
Initially this implemantion could be placed in the [odh-model-controller](https://github.com/opendatahub-io/odh-model-controller) repository.

**Describe the solution you'd like**
Implement the logic described in [this diagram](https://github.com/opendatahub-io/model-registry/issues/104#issuecomment-1782358019), focusing on the `odh-model-controller` actor specified in the sequence diagram.

**Describe alternatives you've considered**
n/a

**Additional context**
Splitting this issue into multiple tasks:
- [ ] https://github.com/opendatahub-io/model-registry/issues/174
- [ ] https://github.com/opendatahub-io/model-registry/issues/175

PR #200, opened 2023-11-23
Override not editable fields during updates

### LINKED ISSUE #199: Unable to update InferenceService and ServeModel
**Describe the bug**
Unable to update `InferenceService` and `ServeModel` through the **proxy**, getting:

```
{"code":"","message":"strconv.Atoi: parsing \"\": invalid syntax"}

```

The issue is because going through the proxy user provide `*Update` entities that do not have NOT editable fields, e.g., `RegisteredModelId` for InferenceService - the core layer is expecting it as this is a required field for `InferenceService` objects.

**To Reproduce**
1. Create a new InferenceService
2. Try to update it using curl
```
url -X 'PATCH' \
  'http://localhost:8080/api/model_registry/v1alpha1/inference_services/5' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "modelVersionId": "3",
  "runtime": "model-server",
  "state": "UNDEPLOYED"
}'
```
3. See error
```
{"code":"","message":"strconv.Atoi: parsing \"\": invalid syntax"}

```

**Expected behavior**
All NOT EDITABLE fields should be automatically filled during the update.

**Additional context**
n/a

lampajr:
Fixes https://github.com/opendatahub-io/model-registry/issues/199

## Description

As per related issue description, this PR aims to fix the update issue for `InferenceService` and `ServeModel`.
Created `OverrideNotEditableFor<Entity>` converters that take `<existing, update>` and override the NOT editable fields in the `update` object by taking those values from the `existing`.

With this approach I was able to remove the `Name` override in the core functions.

**NOTE**: everytime we add a new field in the openapi spec that can be editable, we should `ignore` it in the `OverrideNotEditableFor<Entity>` converters.

> Since I have updated the `goverter` version, I recommend to run `make clean/deps` and then `make build`  to download te proper version of goverter.

## How Has This Been Tested?

```
make test
```

## Merge criteria:

### DIFF
--- a/internal/converter/openapi_converter_test.go
+++ b/internal/converter/openapi_converter_test.go
@@ -1,0 +1,160 @@
+package converter
+
+import (
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"reflect"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/opendatahub-io/model-registry/pkg/openapi"
+)
+
+// visitor
+type visitor struct {
+	t        *testing.T
+	entities map[string]*oapiEntity
+}
+
+func newVisitor(t *testing.T, f *ast.File) visitor {
+	return visitor{
+		t: t,
+		entities: map[string]*oapiEntity{
+			"RegisteredModel": {
+				obj: openapi.RegisteredModel{},
+			},
+			"ModelVersion": {
+				obj: openapi.ModelVersion{},
+			},
+			"ModelArtifact": {
+				obj: openapi.ModelArtifact{},
+			},
+			"ServingEnvironment": {
+				obj: openapi.ServingEnvironment{},
+			},
+			"InferenceService": {
+				obj: openapi.InferenceService{},
+			},
+			"ServeModel": {
+				obj: openapi.ServeModel{},
+			},
+		},
+	}
+}
+
+func (v *visitor) extractGroup(regex *regexp.Regexp, s string) string {
+	extracted := regex.FindStringSubmatch(s)
+	if len(extracted) != 2 {
+		v.t.Errorf("unable to extract groups from %s for %s", regex.String(), s)
+	}
+	// the first one is the wole matched string, the second one is the group
+	return extracted[1]
+}
+
+func (v *visitor) getEntity(name string) *oapiEntity {
+	val, ok := v.entities[name]
+	if !ok {
+		v.t.Errorf("openapi entity not found in the entities map: %s", name)
+	}
+	return val
+}
+
+func (v visitor) Visit(n ast.Node) ast.Visitor {
+	if n == nil {
+		return nil
+	}
+
+	switch d := n.(type) {
+	case *ast.InterfaceType:
+		for _, m := range d.Methods.List {
+			methodName := m.Names[0].Name
+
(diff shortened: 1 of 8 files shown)


tarilabs (dismissed):
Great catch! Thank you so much

As discussed offline, if there is a quick way via introspection/parsing to determine the `// annotation` so to make sure is updated with the list of fields using pointer semantic in the struct, it would be awesome.
If that takes too much time, could be postponed for later

lampajr:
> As discussed offline, if there is a quick way via introspection/parsing to determine the `// annotation` so to make sure is updated with the list of fields using pointer semantic in the struct, it would be awesome. If that takes too much time, could be postponed for later

+1 I will try to investigate if there is an easy way to ensure that all fields that are editable are NOT overridden, keep you posted here.

lampajr:
> +1 I will try to investigate if there is an easy way to ensure that all fields that are editable are NOT overridden, keep you posted here.

I think I have found a possible approach for testing this using reflection and inspection of the goverter methods directives. Hope to provide a proposal on Monday :)

internal/converter/openapi_converter_test.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+		v.t.Errorf("missing fields to be ignored for OverrideNotEditableFor* goverter methods:\n%v", missingFieldsMsg)
+	}
+}
+
+// checkEntity check if there are missing fields to be ignored in the override method
+func checkEntity(entity *oapiEntity) []string {
+	res := []string{}
+	objType := reflect.TypeOf(entity.obj)
+	for i := 0; i < objType.NumField(); i++ {
```
tarilabs, line 124: I suspect there must be a way to make Field(s) iterable with `range` but it's not important for now

internal/converter/openapi_converter_test.go
```diff
@@ -0,0 +1,160 @@
...
+	switch d := n.(type) {
+	case *ast.InterfaceType:
+		for _, m := range d.Methods.List {
+			methodName := m.Names[0].Name
+
+			if converterMethodPattern.MatchString(methodName) {
+				entityName := v.extractGroup(converterMethodPattern, methodName)
+				entity := v.getEntity(entityName)
+				// there should be just one doc comment matching ignoreDirectivePattern
+				for _, c := range m.Doc.List {
```
tarilabs, line 80: I understand this extract all code comments pertaining to Methods. Nice, just as expected also from JavaParser AST I believe structure is analogous

tarilabs (approved):
Awesome! I understand works as expected, and the parser used is even in Go stdlib

internal/converter/openapi_converter_test.go
lampajr: TBH I couldn't find a different way to iterate over those fields, at least using the `reflect` pkg and types. But yeah we can figure out if we can improve it in a later iteration

internal/converter/openapi_converter_test.go
lampajr: yeah correct, it extracts all of them and then I just had to filter the one we are insterested in.

lampajr:
Going to merge this to unblock the development of https://github.com/opendatahub-io/model-registry/issues/173

PR #209, opened 2023-11-28
Fix python client MR type definitions

### LINKED ISSUE #206: MR python client must not expect ModelVersion or ModelArtifact in superentity call
**Describe the bug**
When RegisteredModel miss ModelVersion or a ModelVersion miss ModelArtifact, an error throws.

**To Reproduce**
given `mr` as model registry python client
1. Create a RegisteredModel with upsert call
2. Call `mr.get_registered_model_by_params(name=registeredmodel_name)` where `registeredmodel_name` is the name used in the previous call

image

**Expected behavior**
If a RegisteredModel is missing any ModelVersion, should not throw error on the user, but should be managed.
If a ModelVersion is missing any ModelArtifact, should not throw error on the user, but should be managed.

**Additional context**
if a previous call failed in-between upserting ModelVersion or ModelArtifact, the end result is that the chained superentity is now unusable (as you can no longer retrieve it)
For example, you successfully create RM, but for some reason MV creation fails; nor MR is no longer retrievable.

antisaling:
## Description
Makes ModelVersion and RegisteredModel simpler by removing dependencies on related types. Note that MV still depends on BaseArtifact, but this mapping is not yet done as the user still can't push those objects to the store.

Tests were updated to match the new implementation. In the end, the code is simplified.

Closes: #206

## Merge criteria:

### DIFF
--- a/clients/python/README.md
+++ b/clients/python/README.md
@@ -70,9 +70,6 @@ another_model = registry.get_registered_model_by_id("another-model-id")
 
 another_version = registry.get_model_version_by_id("another-version-id", another_model.id)
 
-# fetching a version will also fetch its associated model artifact
-model_artifact = another_version.model
-
 another_trained_model = registry.get_model_artifact_by_id("another-model-artifact-id")
 ```
 
--- a/clients/python/README.md
+++ b/clients/python/README.md
@@ -68,9 +68,6 @@ To query objects using IDs, do
 ```py
 another_model = registry.get_registered_model_by_id("another-model-id")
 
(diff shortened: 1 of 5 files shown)


tarilabs (approved):
looks to me aligns with behaviour from REST/go side; thank you!

@lampajr I would appreciate also your pair of eyes on this one, please

lampajr (approved):
Looks good to me! 
I see test cases already cover the use case described in the issue, thx a lot @isinyaaa !

Issue #129, opened 2023-11-07
Create higher level Python APIs for user


rareddy:
**Is your feature request related to a problem? Please describe.**
Python APIs are currently defined at par with GO APIs that are more focused on the lower level logical layer of the Model Registry. 

**Describe the solution you'd like**
For user usability purposes, to be used in the Notebooks and Pipeline components should encapsulate these lower-level APIs into functional calls like `register_model` or `list_registered_models` etc.   First, figure out the API and define it here in the issue. once there is acceptance on API then follow up with implementation and testing of the API

**Additional context**
Consider looking at other popular open source model registries to get inspiration about the type of calls to expose

antisaling:
## First iteration proposal

As raised by @tarilabs on #158, the Python client doesn't have a strict mapping with the Go layer yet. This should be fixed so we can introduce a new abstraction layer for users of the Python client.

Having that in mind I'll propose that we adopt the following API for the first iteration:

```py
from model_registry import ModelRegistry

# these could be set by env variables
registry = ModelRegistry(address="server-address", port="port", author="author")

model = registry.register_model("my-model",
                    "s3://path/to/model",
                    format_name="onnx",
                    format_version="v1",
                    secret_account_name="access",
                    version="v2.0",
                    version_description="lorem ipsum",
                   )

model = registry.get_registered_model("my-model")

version = registry.get_model_version("my-model", "v2.0")

experiment = registry.get_model_artifact("my-model", "v2.0")
```

## Next steps

From what I've seen other clients do, a better interface could be like:

```py
model = registry.models.new("my-model")

version = model.versions.new("v2.0", description="improved inference time")

version.model.new("s3://path/to/model", format_name="onnx", format_version="v1", description="lorem ipsum")
version.tags = ["kind:onnx", "fast"]
version.metadata = {"foo": "bar"}

# version.artifacts.new(...)
```

Please let me know your opinion on this.

rareddy:
What does Python language patterns say about or which one is a common pattern between two?

antisaling:
> What does Python language patterns say about or which one is a common pattern between two?

The second pattern is more intuitive, but its semantics are centered around REST APIs -- think that you'd try to mimic REST routes using classes, and the endpoints are methods.

While REST might be a little cleaner from my perspective, this is very different from `mlflow`'s approach, too, so I'm not entirely sure on the best approach for the MR client.

rareddy:
[...]

PR #210, opened 2023-11-28
Introduce higher-level API for py client

antisaling:
## Description

Introduce the first iteration of a higher-level Python API as proposed on #129.

Related: #129
Depends on: #209

## How Has This Been Tested?

New tests were introduced on `tests/test_client.py`.

## Merge criteria:

### DIFF
--- a/clients/python/src/model_registry/__init__.py
+++ b/clients/python/src/model_registry/__init__.py
@@ -2,11 +2,8 @@
 
 __version__ = "0.1.0"
 
-from .client import ModelRegistry
-from .types import ListOptions, OrderByField
+from ._client import ModelRegistry
 
 __all__ = [
     "ModelRegistry",
-    "ListOptions",
-    "OrderByField",
 ]
--- a/clients/python/src/model_registry/_client.py
+++ b/clients/python/src/model_registry/_client.py
@@ -1,0 +1,148 @@
+"""Standard client for the model registry."""
+from __future__ import annotations
+
+from .core import ModelRegistryAPIClient
+from .exceptions import StoreException
+from .types import ModelArtifact, ModelVersion, RegisteredModel
+
+
+class ModelRegistry:
+    """Model registry client."""
+
+    def __init__(
+        self,
+        server_address: str,
+        port: int,
+        author: str,
+        client_key: str | None = None,
+        server_cert: str | None = None,
+        custom_ca: str | None = None,
+    ):
+        """Constructor.
+
+        Args:
+            server_address (str): Server address.
+            port (int): Server port.
+            author (str): Name of the author.
+            client_key (str, optional): The PEM-encoded private key as a byte string.
+            server_cert (str, optional): The PEM-encoded certificate as a byte string.
+            custom_ca (str, optional): The PEM-encoded root certificates as a byte string.
+        """
+        # TODO: get args from env
+        self._author = author
+        self._api = ModelRegistryAPIClient(
+            server_address, port, client_key, server_cert, custom_ca
+        )
+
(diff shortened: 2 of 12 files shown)


clients/python/src/model_registry/_client.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def _register_model_artifact(
+        self, mv: ModelVersion, uri: str, /, **kwargs
+    ) -> ModelArtifact:
+        assert mv.id is not None, "Model version must have an ID"
+        ma = ModelArtifact(mv.model_name, uri, **kwargs)
+        self._api.upsert_model_artifact(ma, mv.id)
+        return ma
+
+    def register_model(
```
tarilabs, line 65: should we rename this ~"register model artifact" or something else to better indicate this is performing the whole chain of inserting (or reusing) RegisteredModel, ModelVersion and ModelArtifact? 🤔

tarilabs (dismissed):
minor comment which I believe is better to address before merging, but I don't want to hold merging solely based on that.

Thank you so much this will be extremely helpful

clients/python/src/model_registry/_client.py
lampajr: In this regard IMO `register_model` is good enough, I found the same pattern on [mlflow api](https://mlflow.org/docs/latest/model-registry.html#api-workflow):
- `mlflow.register_model` high level api to register the full model (model, version and artifact)
- `client.create_registered_model` raw api to create empty model - the same as our `upsert_registered_model`
- `client.create_model_version` raw api to create a new model version - the same as our `upsert_model_version`

wdyt?

clients/python/src/model_registry/__init__.py
```diff
@@ -2,11 +2,8 @@
 
 __version__ = "0.1.0"
 
-from .client import ModelRegistry
-from .types import ListOptions, OrderByField
+from ._client import ModelRegistry
```
lampajr, line 5: So if I correctly understood, with this we are going to "export" just the high level API and not the low level one (i.e., `ModelRegistryAPIClient`) - this way users cannot directly use the underlying API, right?

If I am right, is this what we want? I am thinking how user could update for instance a model artifact metadata if we are not exposing the underlying api.

If I am not right, forget what I said :)

clients/python/src/model_registry/_client.py
tarilabs: yes, thank you for making me notice that, lgtm 👍

clients/python/src/model_registry/__init__.py
antisaling: Yes, by default we only export the higher-level API, but the user can also do `from model_registry.core import ModelRegistryAPIClient as MRAPI` and use the lower-level bindings the same way they would in go.

Edit: clarification

> If I am right, is this what we want? I am thinking how user could update for instance a model artifact metadata if we are not exposing the underlying api.

I'm actually planning on another more powerful version of this higher-level API, but if it's really important that this is exposed by default, we can think of workarounds :)

lampajr: Well actually if the user can import the lower level api that is fine for me, make sense to export by default just the higer level one.

But we should guarantee a way to make use of the lower level if the user want to use it, at least for now I think

tarilabs: > But we should guarantee a way to make use of the lower level if the user want to use it, at least for now I think

+1

Thanks for the clarifications Isabella

tarilabs (approved):

lampajr (approved):

PR #213, opened 2023-11-29
Improve core testing using single test container

### LINKED ISSUE #212: Use single TestContainer for core testing
**As Is:**
Right now core testing startup and teardown one TestContainer per single test case, this is introducing an overhead that could make `core_test` running longer.

**Motivation of the change:**
Thanks to the fact that if we remove the metadata sqlite file used by mlmd, the server will automatically recreate it as soon as it is needed, we can have just one single MLMD test container

**To Be:**
Setup a single MLMD TestContainer _before all_ test cases of `core_test`, then _after each_ test case remove the metadata sqlite file.

lampajr:
Fixes https://github.com/opendatahub-io/model-registry/issues/212

## Description

Refactor the core testing (file `core_test`) to make use of a single MLMD TestContainer, thanks to this improvement we should reduce the overhead caused by starting up and tearding down the test container itself.

From a technical point of view I made use of testify test `Suite` object, in this way `before/after each` methods are much more explicit and they do not have to be called by every test method.

## How Has This Been Tested?

```bash
make test
```

## Merge criteria:

### DIFF
--- a/pkg/core/core_test.go
+++ b/pkg/core/core_test.go
@@ -10,7 +10,7 @@ import (
 	"github.com/opendatahub-io/model-registry/internal/testutils"
 	"github.com/opendatahub-io/model-registry/pkg/api"
 	"github.com/opendatahub-io/model-registry/pkg/openapi"
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
 	"google.golang.org/grpc"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/core/core_test.go
```diff
@@ -45,12 +45,26 @@ var (
...
+	defer teardown(t)
+
+	coreTestSuite := CoreTestSuite{
+		grpcConn:   grpcConn,
+		mlmdClient: mlmdClient,
 	}
+	suite.Run(t, &coreTestSuite)
+}
 
+// befor each test case
```
tarilabs:
```suggestion
// before each test case
```

tarilabs (dismissed):
thank you very much!

pkg/core/core_test.go
lampajr: ops typo :smile:

tarilabs (approved):

PR #220, opened 2023-11-30
Improve err mgmt when providing wrong entity

### LINKED ISSUE #217: Improve error management when providing wrong entity
Improve the error management whenever the wrong entity is passed in any call, i.e., whenever the user provide the parent resource Id of an entity that is different from the expected one. 

In these cases the returned error is something like:
```
"invalid TypeId, expected 11 but received 10"
```

**How to reproduce**

Startup the model registry:
```bash
docker compose -f docker-compose.yaml up
```

Create a registered model:
```bash
curl -X 'POST' \
  'http://localhost:8080/api/model_registry/v1alpha1/registered_models' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "name": "MyModel",
  "state": "LIVE"
}'
```

Create model version:
```bash
curl -X 'POST' \
  'http://localhost:8080/api/model_registry/v1alpha1/model_versions' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "name": "v1",
  "state": "LIVE",
  "registeredModelID": "1"
}'
```

Try to create a model artifact by wrongly passing the registered model id (i.e., `1`) instead of the model version one (i.e., `2`):
```bash
curl -X 'POST' \
  'http://localhost:8080/api/model_registry/v1alpha1/model_versions/1/artifacts' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "state": "UNKNOWN",
  "name": "v1-artifact",
  "artifactType": "model-artifact"
}'
```

You will get the error:

```bash
{
  "message": "invalid TypeId, expected 11 but received 10"
}
```

The error itself is right as the mapper is expecting a `ModelVersion` (type id `11` in this case) but it received a `RegisteredModel` (type id `10`) but it is not so easy to understand for external users, furthermore those type IDs are "internal" and do not have any meaning outside MLMD.

My expectation would be to return an error like `invalid entity, expected ModelVersion but received RegisteredModel as parent resource. please check the provided id`.

To fix this we should keep track (internally) of the mapping between the MLMD type IDs and the corresponding entity name, this way it would be easy to provide a meaningful error msg.

n/a

**Additional context**
n/a

lampajr:
Fixes https://github.com/opendatahub-io/model-registry/issues/217

## Description

As per linked issue, I managed to keep track of the maaping between the internal MLMD typedId and the corresponding name (avoiding to perform additional request on MLMD just to retrieve the name).

Following the "How to reproduce" steps in the issue, the final error result is:
```bash
"message":"invalid entity: expected odh.ModelVersion but received odh.RegisteredModel, please check the provided id"
```

I think this could be much more helpful/useful from user perspective.

In addition to this specific err mgmt I added:
* Moved odh MLMD types in a separated `constants.go` file, easier to keep track of them.
* Improved the `mapper` coverage

## How Has This Been Tested?

```bash
make test-cover
```

## Merge criteria:

### DIFF
--- a/internal/converter/mlmd_converter_util_test.go
+++ b/internal/converter/mlmd_converter_util_test.go
@@ -624,21 +624,18 @@ func TestMapInferenceServiceProperties(t *testing.T) {
 	assertion.Equal(int64(3), props["serving_environment_id"].GetIntValue())
 
 	// serving and model id must be provided and must be a valid numeric id
-	props, err = MapInferenceServiceProperties(&openapi.InferenceService{})
+	_, err = MapInferenceServiceProperties(&openapi.InferenceService{})
 	assertion.NotNil(err)
 	assertion.Equal("missing required RegisteredModelId field", err.Error())
-	assertion.Equal(0, len(props))
 
-	props, err = MapInferenceServiceProperties(&openapi.InferenceService{RegisteredModelId: "1"})
+	_, err = MapInferenceServiceProperties(&openapi.InferenceService{RegisteredModelId: "1"})
 	assertion.NotNil(err)
 	assertion.Equal("missing required ServingEnvironmentId field", err.Error())
-	assertion.Equal(0, len(props))
 
 	// invalid int
-	props, err = MapInferenceServiceProperties(&openapi.InferenceService{RegisteredModelId: "aa"})
+	_, err = MapInferenceServiceProperties(&openapi.InferenceService{RegisteredModelId: "aa"})
 	assertion.NotNil(err)
 	assertion.Equal("invalid numeric string: strconv.Atoi: parsing \"aa\": invalid syntax", err.Error())
-	assertion.Equal(0, len(props))
 }
 
 func TestMapServeModelType(t *testing.T) {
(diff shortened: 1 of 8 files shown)


lampajr:
Overall coverage improved: Comparison is base [(ee64553)](https://app.codecov.io/gh/opendatahub-io/model-registry/commit/ee6455355f140e3843b56bd17e4185a4e7551b7a?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=opendatahub-io) 71.82% compared to head [(b33a61e)](https://app.codecov.io/gh/opendatahub-io/model-registry/pull/220?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=opendatahub-io) 83.39%.

internal/converter/mlmd_converter_util_test.go
```diff
@@ -625,6 +626,18 @@ func TestMapInferenceServiceProperties(t *testing.T) {
 	// serving and model id must be provided and must be a valid numeric id
 	props, err = MapInferenceServiceProperties(&openapi.InferenceService{})
 	assertion.NotNil(err)
+	assertion.Equal("missing required RegisteredModelId field", err.Error())
+	assertion.Equal(0, len(props))
```
tarilabs: up to you, but I typically prefer "unspecified behaviour" for the left tuple value; ie I'd rewrite the call as `_, err := ...` this way we will never receive "complains" of "I saw your test specified the left zero-value in case of error" by user of the API :)

tarilabs (dismissed):
minor comment for your evaluation; Thank you so much for taking up this chore, improving error msg is always beneficial for DX and UX !

internal/converter/mlmd_converter_util_test.go
lampajr: I definitely agree with you, change applied with commit https://github.com/opendatahub-io/model-registry/pull/220/commits/06c49e1b7198a233819e206a0d7b89e96de64e15

tarilabs (approved):

PR #225, opened 2023-12-04
py: using Testcontainers and test strategy as go

### LINKED ISSUE #224: MR python client test to use Testcontainers (for Python)
Currently python tests rely on the embedded compiled binary mlmd server inside the platform/architecture specific python wheel.

Using Testcontainers for Python and testing analogously to what we done as testing strategy on go Core layer 

Unfortunately this would make it unavailable in docker-in-docker setups such as DevContainers. While this problem can be circumvented by manually raising a MLMD server in the host docker, the problem appears to be slowness in FS reactivity when deleting the sqlite file. Considering the ideal solution is likely more about avoiding emulation and rely on gRPC, I would conclude this is still the better solution.

**Additional context**
n/a

tarilabs:
Resolves #224

## Description
Using Testcontainers for Python and testing analogously to what we done as testing strategy on go Core layer 

### Caveats
Unfortunately this would make it unavailable in docker-in-docker setups such as DevContainers. While this problem can be circumvented by manually raising a MLMD server in the host docker, the problem appears to be slowness in FS reactivity when deleting the sqlite file. Considering the ideal solution is likely more about avoiding emulation and rely on gRPC (see recent document submitted to the group), I would conclude this PR and Testcontainers based approach is still the better solution.

## How Has This Been Tested?
```
pytest -v -x -s
```
✅ `40 passed in 14.07s`

Since I'm on M-chip I've used the .toml wheel using only grpc connection. I'd appreciate some 👀 from some other contributors using Intel: @lampajr can you give it a spin when possible, please? @isinyaaa can you give it a spin from the remote server when possible, please?

## Merge criteria:


tarilabs:
edit: Correction, seems like `pytest --cov --cov-config=pyproject.toml --cov-report=xml` from `nox` in GitHub Action is not accounting `testcontainers` despite being listed in the dev dependency from toml 🤔 
@isinyaaa @dtrifiro can you kindly check the GHA setup accordingly, please?

clients/python/pyproject.toml
```diff
@@ -10,6 +10,10 @@ readme = "README.md"
 python = ">= 3.9, < 3.11"
 attrs = "^21.0"
 ml-metadata = "^1.14.0"
+# Testcontainers for python, remote mlmd grpc tested with:
+# ml-metadata = { url = "https://github.com/tarilabs/ml-metadata/releases/download/1.14.0/ml_metadata-1.14.0-py3-none-any.whl" }
```
tarilabs, line 14: 🔍 ☝️  when mentioned:

> Since I'm on M-chip I've used the .toml wheel using only grpc connection. I'd appreciate some 👀 from some other contributors using Intel: @lampajr can you give it a spin when possible, please? @isinyaaa can you give it a spin from the remote server when possible, please?

clients/python/pyproject.toml
```diff
@@ -22,6 +26,7 @@ pytest-cov = "^4.1.0"
 sphinx-autobuild = "^2021.3.14"
 ruff = "^0.1.6"
 mypy = "^1.7.0"
+testcontainers = "^3.7.1"
```
dtrifiro, line 29: Looking at the fixture implementation, it looks kinda clutterered. How about trying https://pypi.org/project/pytest-docker/?

tarilabs: That's a possibility; the reason for adopting TestContainers is that we're using it already on the Go side, hence for alignment.

In any case, this is discussion we can have separately, currently my main doubt here is as summarized in: https://github.com/opendatahub-io/model-registry/pull/225#issuecomment-1838576589

lampajr:
> edit: Correction, seems like `pytest --cov --cov-config=pyproject.toml --cov-report=xml` from `nox` in GitHub Action is not accounting `testcontainers` despite being listed in the dev dependency from toml 🤔 @isinyaaa @dtrifiro can you kindly check the GHA setup accordingly, please?

IIUC it should be enough to change https://github.com/opendatahub-io/model-registry/blob/0621864cea0d438df528afd1aac0bd6eb39fc0c2/clients/python/noxfile.py#L52
to 
```python
 session.install("coverage[toml]", "pytest", "pytest-cov", "pygments", "testcontainers") 
```

clients/python/pyproject.toml
lampajr: fwiw I tried all three approaches, and all of them worked (all tests green) on Intel :rocket:

tarilabs:
Edit: codecov now working

lampajr (dismissed):
lgtm

clients/python/tests/conftest.py
```diff
@@ -7,18 +7,71 @@
...
+
+    time.sleep(3) # allowing some time for mlmd grpc to fully stabilize (is "spent" once per pytest session anyway)
+    store = metadata_store.MetadataStore(cfg)
+    wait_for_grpc(container, store)
+
+    return MLMDStore(cfg) 
+
+
+def model_registry_root(request):
+    return request.config.rootdir + '/../..'
```
antisaling: I don't think this function is necessary, I also didn't find `rootdir` in their API docs so I'd use the safer [`rootpath`](https://docs.pytest.org/en/latest/reference/reference.html#pytest.Config.rootpath).

antisaling: Per the previous comment, as `rootpath` is a `Path` object, it's possible to do:

```suggestion
def model_registry_root(request):
    return (request.config.rootpath / '../..').resolve()  # resolves to absolute path
```

you can use the same trick on other parts as well, printing this will also be fine :)

clients/python/tests/conftest.py
```diff
@@ -7,18 +7,71 @@
...
-@pytest.fixture()
-def plain_wrapper() -> MLMDStore:
-    config = ConnectionConfig()
-    config.fake_database.SetInParent()
-    return MLMDStore(config)
+@pytest.fixture(scope='session')
+def plain_wrapper(request) -> MLMDStore:
+    print("Assuming this is Model Registry clients/python directory:", request.config.rootdir)
+    model_registry_root_dir = model_registry_root(request)
+    print("Assuming this is the Model Registry root directory:", model_registry_root_dir)
```
antisaling, line 25: nit: we could skip this log

clients/python/tests/conftest.py
```diff
@@ -7,18 +7,71 @@
...
+    sqlite_db_file = model_registry_root_dir + '/test/config/ml-metadata/metadata.sqlite.db'
+    if os.path.exists(sqlite_db_file):
+        raise FileExistsError(f"The file {sqlite_db_file} already exists; make sure to cancel it before running these tests.")
```
antisaling: Can't we use a temporary file for the DB?

antisaling (changes requested):
Thanks for the change Matteo :)

I noticed we haven't enabled linting on tests, so I'd like to ask you to wait before merging this and then rebase on top of #231 when that's merged so we can iron out some things before merging. Is that OK?

clients/python/pyproject.toml
antisaling: I agree that `pytest-docker` would probably lead to a simpler solution, is there any special reason we want to align both implementations on this?

tarilabs: As mentioned, this brings alignment on Integration Testing strategy and "infrastructure" if compared to Go side.

Another thing which later did not find in `pytest-docker` is ryuk container orchestration, which is typically helpful when performing test locally. But please let me know if I missed.

I've tested on M2 Mac, this works; Andrea tested on Intel works; on GHA is green. Is there something not working on someone else platform motivating a requested-change on the (allegedly already-working) solution presented in this PR which I missed, please? 🤔

tarilabs:
> I noticed we haven't enabled linting on tests, so I'd like to ask you to wait before merging this and then rebase on top of #231 when that's merged so we can iron out some things before merging. Is that OK?

sure, noted. Will hold this merge to rebase on #231

clients/python/tests/conftest.py
tarilabs: We can evaluate this in a subsequent PR since we're also making analogous consideration on IT infra on Go side. We should as well consider using cmd line option in place of the pb config file then.

clients/python/tests/conftest.py
tarilabs: you mean line 28 or simply all the prints? Those were helpful during debugging and by default I understand they are suppressed unless you launch with `-s`. Do you want to use an alternative logging framework/call?

clients/python/tests/conftest.py
tarilabs: per https://github.com/opendatahub-io/model-registry/pull/225/files/09c0674ced24aa2a45d3eaaa28c905667115a543#diff-31172aa4bf9754641d05923f27755c9b5886c54e7e74ad170ab81b940e580d7e

tarilabs:
requesting review after rebasing per original request in https://github.com/opendatahub-io/model-registry/pull/225#pullrequestreview-1766186790

clients/python/tests/conftest.py
antisaling: I mean the log on line 26, as it's just `../../` from the last one.

clients/python/tests/conftest.py
```diff
@@ -1,24 +1,76 @@
...
-    config.fake_database.SetInParent()
-    return MLMDStore(config)
+# ruff: noqa: PT021 supported
+@pytest.fixture(scope="session")
+def plain_wrapper(request) -> MLMDStore:
+    print("Assuming this is Model Registry clients/python directory:", request.config.rootdir)
+    model_registry_root_dir = model_registry_root(request)
+    print("Assuming this is the Model Registry root directory:", model_registry_root_dir)
+    sqlite_db_file = f"{model_registry_root_dir}/test/config/ml-metadata/metadata.sqlite.db"
+    if os.path.exists(sqlite_db_file):
```
antisaling:
```suggestion
    sqlite_db_file = model_registry_root_dir / "test/config/ml-metadata/metadata.sqlite.db"
    if sqlite_db_file.exists():
```

clients/python/tests/conftest.py
```diff
@@ -1,24 +1,76 @@
...
+    cfg = MetadataStoreClientConfig()
+    cfg.host = "localhost"
+    cfg.port = int(container.get_exposed_port(8080))
+    print(cfg)
+
+    # this callback is needed in order to perform the container.stop()
+    # removing this callback might result in mlmd container shutting down before the tests had chance to fully run,
+    # and resulting in grpc connection resets.
+    def teardown():
+        os.system(f"rm {model_registry_root_dir}/test/config/ml-metadata/metadata.sqlite.db") # noqa governed test
```
antisaling:
```suggestion
        os.system(f"rm {sqlite_db_file}") # noqa governed test
```
?

clients/python/tests/conftest.py
```diff
@@ -1,24 +1,76 @@
...
+
+@pytest.fixture(autouse=True)
+def _plain_wrapper_after_each(request, plain_wrapper: MLMDStore):
+    model_registry_root_dir = model_registry_root(request)
+    sqlite_db_file = f"{model_registry_root_dir}/test/config/ml-metadata/metadata.sqlite.db"
+    def teardown():
+        os.system(f"rm {sqlite_db_file}") # noqa governed test
+        print("plain_wrapper_after_each done.")
+
+    request.addfinalizer(teardown)
```
antisaling: can you clarify the purpose of this autouse fixture?

tarilabs: Sure.
After each test, remove the SQLite file.
Analogously how we done on Go side with @lampajr

clients/python/tests/conftest.py
tarilabs: but that I found helpful with `-s` to make sure it was pointing to the expected directory?
I can remove it, but personally I found it helpful (considering it is displayed only with opt-in with `-s` on test launch).

clients/python/tests/conftest.py
tarilabs: After each test *using the `plain_wrapper`, to clarify

dtrifiro: Why not use a fixture to provide the sqlite db and then do the teardown after the yield?

tarilabs: > Why not use a fixture to provide the sqlite db and then do the teardown after the yield?

unless I missed something, that's exactly what the lines
https://github.com/opendatahub-io/model-registry/pull/225/files#diff-31172aa4bf9754641d05923f27755c9b5886c54e7e74ad170ab81b940e580d7eR22-R23

above are doing.

With the difference the provision is session scope.
While the removal of the SQLite file is done after each test (using the plain_wrapper).

This allow us to have the Container spin up once, while a new SQLite file is created by the MLMD server at each test.

This approach also on Go layer allowed us to save some time if compared to shutdown and reprovision of the MLMD container.

clients/python/tests/conftest.py
tarilabs: ok I see better now, we could remove line 24

clients/python/tests/conftest.py
```diff
@@ -1,24 +1,76 @@
...
-def plain_wrapper() -> MLMDStore:
-    config = ConnectionConfig()
-    config.fake_database.SetInParent()
-    return MLMDStore(config)
+# ruff: noqa: PT021 supported
+@pytest.fixture(scope="session")
+def plain_wrapper(request) -> MLMDStore:
+    print("Assuming this is Model Registry clients/python directory:", request.config.rootdir)
+    model_registry_root_dir = model_registry_root(request)
+    print("Assuming this is the Model Registry root directory:", model_registry_root_dir)
```
tarilabs:
```suggestion
    model_registry_root_dir = model_registry_root(request)
    print("Assuming this is the Model Registry root directory:", model_registry_root_dir)
```

clients/python/tests/conftest.py
tarilabs: noted in https://github.com/opendatahub-io/model-registry/pull/225/files#r1422653146

tarilabs: done with https://github.com/opendatahub-io/model-registry/pull/225/commits/845f5c7c172e0f0eb1abb52db56056c8a63d45d3

clients/python/tests/conftest.py
dtrifiro: How about using the `tmp_dir` fixture to provide a new workspace for each test?

tarilabs: > How about using the `tmp_dir` fixture to provide a new workspace for each test?

Yes, that would a possible future improvement also on Go side, but we can do that in subsquent PRs as mentioned in https://github.com/opendatahub-io/model-registry/pull/225#discussion_r1417271106

clients/python/pyproject.toml
antisaling: I don't know if I missed something but I keep getting this prompt after each test when running with `-s`:
image

clients/python/tests/conftest.py
```diff
@@ -1,24 +1,75 @@
...
+    sqlite_db_file = model_registry_root_dir / "test/config/ml-metadata/metadata.sqlite.db"
+    if sqlite_db_file.exists():
+        msg = f"The file {sqlite_db_file} already exists; make sure to cancel it before running these tests."
+        raise FileExistsError(msg)
+    container = DockerContainer("gcr.io/tfx-oss-public/ml_metadata_store_server:1.14.0")
+    container.with_exposed_ports(8080)
+    container.with_volume_mapping(f"{model_registry_root_dir}/test/config/ml-metadata/", "/tmp/shared", "rw") # noqa this is file target in container
+    container.with_env("METADATA_STORE_SERVER_CONFIG_FILE", "/tmp/shared/conn_config.pb") # noqa this is target in container
```
antisaling: nit:

```suggestion
    shared_volume = model_registry_root_dir / "test/config/ml-metadata"
    sqlite_db_file = shared_volume / "metadata.sqlite.db"
    if sqlite_db_file.exists():
        msg = f"The file {sqlite_db_file} already exists; make sure to cancel it before running these tests."
        raise FileExistsError(msg)
    container = DockerContainer("gcr.io/tfx-oss-public/ml_metadata_store_server:1.14.0")
    container.with_exposed_ports(8080)
    container.with_volume_mapping(shared_volume, "/tmp/shared", "rw") # noqa this is file target in container
    container.with_env("METADATA_STORE_SERVER_CONFIG_FILE", "/tmp/shared/conn_config.pb") # noqa this is target in container
```

clients/python/pyproject.toml
tarilabs: that's very strange, never got them 🤔 
@lampajr would you mind retest again this PR, using `poetry run test -x -s` please?

clients/python/tests/conftest.py
tarilabs: I see, you want to reuse the `shared_volume` ; sure np

clients/python/tests/conftest.py
```diff
@@ -1,24 +1,76 @@
...
+    container.with_volume_mapping(shared_volume, "/tmp/shared", "rw") # noqa this is file target in container
+    container.with_env("METADATA_STORE_SERVER_CONFIG_FILE", "/tmp/shared/conn_config.pb") # noqa this is target in container
+    container.start()
+    wait_for_logs(container, "Server listening on")
+    os.system('docker container ls --format "table {{.ID}}\t{{.Names}}\t{{.Ports}}" -a') # noqa governed test
+    print("waited for logs and port")
+    cfg = MetadataStoreClientConfig()
+    cfg.host = "localhost"
+    cfg.port = int(container.get_exposed_port(8080))
+    print(cfg)
```
antisaling: sorry for commenting this now, you're too quick!

Nit:

```suggestion
    assert container.get_exposed_port(8080)
    cfg = MetadataStoreClientConfig(host="localhost", port=8080)
```

clients/python/pyproject.toml
antisaling: I believe this can be fixed with a simple `rm -f` on teardown commands :)

clients/python/tests/conftest.py
```diff
@@ -78,3 +130,23 @@ def mr_api(store_wrapper: MLMDStore) -> ModelRegistryAPIClient:
...
+    while True:
+        duration = time.time() - start
+        results = None
+        try:
+            results = store.get_contexts()
+        except errors.UnavailableError as e:
+            print("Container not ready. Retrying...")
+            print(e)
+            stdout = container.get_logs()
+            print(stdout)
```
antisaling: Nit: wouldn't this (or maybe another) order make more sense, as we're retrying after getting the error?

```suggestion
            print(e)
            print("Container logs:\n", container.get_logs())
            print("Container not ready. Retrying...")
```

clients/python/tests/conftest.py
```diff
@@ -1,24 +1,76 @@
...
+    cfg = MetadataStoreClientConfig()
+    cfg.host = "localhost"
+    cfg.port = int(container.get_exposed_port(8080))
+    print(cfg)
+
+    # this callback is needed in order to perform the container.stop()
+    # removing this callback might result in mlmd container shutting down before the tests had chance to fully run,
+    # and resulting in grpc connection resets.
+    def teardown():
+        os.system(f"rm {sqlite_db_file}") # noqa governed test
```
antisaling: Does it make sense to delete this again on container teardown? It should be enough just doing it on the other fixture.

clients/python/tests/conftest.py
```diff
@@ -1,24 +1,76 @@
...
+
+
+def model_registry_root(request):
+    return (request.config.rootpath / "../..").resolve()  # resolves to absolute path
+
+
+@pytest.fixture(autouse=True)
+def _plain_wrapper_after_each(request, plain_wrapper: MLMDStore):
+    model_registry_root_dir = model_registry_root(request)
+    sqlite_db_file = f"{model_registry_root_dir}/test/config/ml-metadata/metadata.sqlite.db"
```
antisaling: nit: let's use path concatenation as above:

```suggestion
    sqlite_db_file = model_registry_root(request) / "test/config/ml-metadata/metadata.sqlite.db
```

clients/python/pyproject.toml
antisaling: I think this is fine, it was just a suggestion. Thanks for pointing out ryuk, I didn't know about it before :).

antisaling: oh btw, a minor consideration is that I found `testcontainers-py` docs really terse.

clients/python/tests/conftest.py
tarilabs: Sorry, I do not think this is right.
The exposed port 8080 gets mapped to a bindable port on the system running the Testcontainer.
So your proposed line 40 would not work?
Am I missing something?

clients/python/tests/conftest.py
tarilabs: I felt it didn't hurt being declarative about session cleanup.

clients/python/pyproject.toml
tarilabs: Here is output on my end _with_ and without `-s`:
image

tarilabs: should be sovled by https://github.com/opendatahub-io/model-registry/pull/225/commits/de399a8fe7a534cfa22aca1259ea550d4334d897

lampajr (dismissed):
Looks great to me.

Tests are working locally with both `pytest` and `pytest -x -s` commands, furthermore using all possible ml-metadata binaries (`{ url = "https://github.com/tarilabs/ml-metadata/releases/download/1.14.0/ml_metadata-1.14.0-py3-none-any.whl" }` and `{ url = "https://github.com/tarilabs/ml-metadata-remote/releases/download/1.14.0/ml_metadata-1.14.0-py3-none-any.whl" }`).

clients/python/tests/conftest.py
antisaling: oh you're right, sorry I misunderstood what you had done here. Resolving...

clients/python/pyproject.toml
antisaling: Some tests don't need the fixture, so they'll print needless errors when autousing it. Wouldn't it make more sense to include this fixture only where we are using plain_wrapper?

image

tarilabs: I haven't found a way to include conditionally the fixture only when `plain_wrapper` is used by other tests.

I'd suggest if we have a quick solution, great let's implement it; otherwise, this message is only displayed with verbose output, is not making any failures, and this PR in the bigger picture solves to avoid depending on a capability of MLMD which is arch specific (and that we don't use "in prod") and aligns with Go core test strategy (using Testcontainers).

antisaling: I made an experiment to demonstrate that it's possible, check it out here: https://github.com/isinyaaa/model-registry/commit/f3d9e6de3a871243ac806537c3ff522f6666fd16

Some of the changes are because of auto-formatting, the main thing was moving `plain_wrapper` to a separate fixture to have the correct scopes for calling `_clean_db`.

tarilabs: @isinyaaa looking at the linked change, makes sense to me.
Separation of concerns, the session-scoped fixture raises the Testcontainer for the MLMD docker container, but provides the connection (not the plain wrapper).

You then have:
```python
plain_wrapper(mlmd_conn: MetadataStoreClientConfig, _clean_db) 
```

which is at each-test scope.

In principle I can see your proposal working; but at that point, wouldn't it make more sense _not_ having the separate `_clean_db` fixture, and just make the finalizer inside the `plain_wrapper` fixture? 🤔 

ie, move the finalizer of _clean_db inside of plain_wrapper.
At this point fixture plain_wrapper produces the MLMDStore as a result,
but also have the finalizer which erases the SQLite file.

Hope I was able to explain myself?

tarilabs: @isinyaaa should be done with ef976e4

antisaling: that was a nice suggestion @tarilabs  :) I was kinda tired when I proposed that, thanks a lot for putting some thought into it!

antisaling (approved):
LGTM!

lampajr (approved):

clients/python/pyproject.toml
tarilabs: > that was a nice suggestion @tarilabs :)

Teamwork ftw  💪 🚀

Issue #130, opened 2023-11-07
Creating a pip installable package for Python APIs


rareddy:
**Is your feature request related to a problem? Please describe.**
We need a way to import Python libraries into a Notebook or Pipeline component

**Describe the solution you'd like**
Make Model Registry Python available as a `pip` based package

**Describe alternatives you've considered**
[Python: Creating a pip installable package](https://betterscientificsoftware.github.io/python-for-hpc/tutorials/python-pypi-packaging/#:~:text=Inside%20that%20package%20directory%2C%20alongside,be%20installed%20and%20become%20importable.)

antisaling:
The package is now published at https://pypi.org/project/model-registry/0.1.0/.

PR #230, opened 2023-12-05
GHA: add workflow to publish python client on PyPI

antisaling:
## Description

This builds and publishes tagged releases of the Python client.

The tags should be prefixed with `py-v`, then followed by the version number.

Related: #130 

## Merge criteria:


.github/workflows/python-release.yml
```diff
@@ -0,0 +1,47 @@
+name: Release Python client
+
+on:
+  push:
+    tags:
+      - py-v*
```
tarilabs, line 6: it's not very clear to me however what is the role of git-tags, if anyway the actual published version is identified by the toml:

https://github.com/opendatahub-io/model-registry/blob/0621864cea0d438df528afd1aac0bd6eb39fc0c2/clients/python/pyproject.toml#L3

tarilabs (commented):
I believe workflow_dispatch is very meaningful so we can manually execute the publish any time required!

.github/workflows/python-release.yml
antisaling: I'm not a fan of git tags myself, tbh. The alternative I found was to use something like
```sh
bash -o pipefail -c "poetry version | cut -d' ' -f2"
```
to get the version for two subsequent commits (e.g. this [version tagging action](https://github.com/salsify/action-detect-and-tag-new-version/blob/master/src/main.ts#L12) that [hypermodern-python cookiecutter](https://github.com/cjolowicz/cookiecutter-hypermodern-python/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/.github/workflows/release.yml#L45) uses). wdyt?

Edit: typo in the command

tarilabs: Whatever works easiest, I'm just dubious in having separate tagging for Go/Python will be maintained?

We can definitely start with this setup and change later 👍 

(I'm assuming if one inadvertently tag as `py-v0.2.0` while having forgotten to update toml file: we just get an error from pypi; so we update the toml, and re-tag)

dtrifiro: The actual version uploaded to PyPi is indeed set by the `pyproject.toml`. I'd add a check to make sure that `pyproject.toml` version matches the git tag before running the release wortkflow.

tarilabs: that's a nice idea +1

antisaling: addressed on [+38-+42](https://github.com/opendatahub-io/model-registry/pull/230/files#diff-08d5d79624884efe962c68a256bee91079ad74acb3b22d61850797f5bf8aae04R38-R42), wdyt?

tarilabs: > addressed on [+38-+42](https://github.com/opendatahub-io/model-registry/pull/230/files#diff-08d5d79624884efe962c68a256bee91079ad74acb3b22d61850797f5bf8aae04R38-R42), wdyt?

I'm not expert in bash scripting, but the overall look of the command and the explanation of the command by LLM matches expectations :) so looks good to me, thank you!

tarilabs (dismissed):

tarilabs (approved):

PR #243, opened 2023-12-13
Change state to desiredState for InferenceService

### LINKED ISSUE #229: InferenceService state terminology
Right now for `InferenceService` entities we do have a state which can have two different possible values `DEPLOYED` and `UNDEPLOYED`. 

Since these entities are intended to be used to describe the `intention to deploy/undeploy a model` and not the deployment state itself, I think that the values could be misleading.

 I think we should slightly change the possible values to properly keep track of correct IS states:
- `READY_TO_DEPLOY`: represents the intention to deploy an InferenceService (this will be considered by the model controller)
- `DEPLOYED`: this is going to be updated by the model controller, just to avoid that IS actually deployed will remain in READY_TO_DEPLOY state.
- `READY_TO_UNDEPLOY`: intention to undeploy a deployed model (considered by model controller)
- `UNDEPLOYED`:  _[optional]_ represents a not deployed model, we could use _nil_ state instead.

n/a

**Additional context**
Model controller operation are still tracked/audited using `ServeModel` entities, one per deployment.

lampajr:
Fixes https://github.com/opendatahub-io/model-registry/issues/229

## Description
As the `state` property in `InferenceService` is an "intention", set by user, to do something we decided to change the property name from `state` to `desiredState`
- [x] Change in openapi specification
- [x] Change in core and mlmd type property

> Change in python client is not needed as Serving interaction are not implemented there yet

> NOTE: since there is NOT a release yet I renamed the property in mlmd, keep in mind that if you have some DBs running with the old schema you might have to delete it and recreate as MLMD does not accept to rename properties. Maybe this is something we could take into account later, i.e, trying to be resilient to schema changes (or just be everytime backward compatible)

## How Has This Been Tested?

Go tests
```bash
make test
```

## Merge criteria:

### DIFF
--- a/api/openapi/model-registry.yaml
+++ b/api/openapi/model-registry.yaml
@@ -1576,7 +1576,7 @@ components:
             runtime:
               description: Model runtime.
               type: string
-            state:
+            desiredState:
               $ref: '#/components/schemas/InferenceServiceState'
     InferenceServiceCreate:
       description: >-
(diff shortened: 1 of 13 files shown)


lampajr:
Errors unrelated to the change, just temporary issue on codecov side: 
```bash
[2023-12-13T21:18:24.816Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
```

tarilabs (approved):

PR #267, opened 2024-01-16
Decouple MLMD types setup from Go core library

### LINKED ISSUE #265: Decouple MLMD types setup from Go core library
Right now the MLMD types setup is performed during the `ModelRegistryService` setup which acts as both internal core layer and importable Go library - therefore, since this could be used by external users, the MLMD types setup should be decoupled from the Go core library setup.

This is a sub-task of https://github.com/opendatahub-io/model-registry/issues/211

Move the MLMD types setup [1] into a dedicated `internal` package.

As a consequence, other changes are required:
1. The REST layer should invoke the moved MLMD types setup
2. The `ModelRegistryService` should just fetch those types (it needs their IDs) assuming they are already existing. 

[1] https://github.com/opendatahub-io/model-registry/blob/53b744622382220d7fd827f52ba74f4d99b0319b/pkg/core/core.go#L49-L163

n/a

**Additional context**
n/a

tarilabs:
resolves #265 

## Merge criteria:

### DIFF
--- a/internal/mlmdtypes/mlmdtypes.go
+++ b/internal/mlmdtypes/mlmdtypes.go
@@ -1,0 +1,144 @@
+package mlmdtypes
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/opendatahub-io/model-registry/internal/apiutils"
+	"github.com/opendatahub-io/model-registry/internal/constants"
+	"github.com/opendatahub-io/model-registry/internal/ml_metadata/proto"
+	"google.golang.org/grpc"
+)
+
+var (
+	registeredModelTypeName    = apiutils.Of(constants.RegisteredModelTypeName)
+	modelVersionTypeName       = apiutils.Of(constants.ModelVersionTypeName)
+	modelArtifactTypeName      = apiutils.Of(constants.ModelArtifactTypeName)
+	servingEnvironmentTypeName = apiutils.Of(constants.ServingEnvironmentTypeName)
+	inferenceServiceTypeName   = apiutils.Of(constants.InferenceServiceTypeName)
+	serveModelTypeName         = apiutils.Of(constants.ServeModelTypeName)
+	canAddFields               = apiutils.Of(true)
+)
+
+// Utility method that created the necessary Model Registry's logical-model types
+// as the necessary MLMD's Context, Artifact, Execution types etc. in the underlying MLMD service
+func CreateMLMDTypes(cc grpc.ClientConnInterface) (map[string]int64, error) {
+	client := proto.NewMetadataStoreServiceClient(cc)
+
+	registeredModelReq := proto.PutContextTypeRequest{
+		CanAddFields: canAddFields,
+		ContextType: &proto.ContextType{
+			Name: registeredModelTypeName,
+			Properties: map[string]proto.PropertyType{
+				"description": proto.PropertyType_STRING,
+				"state":       proto.PropertyType_STRING,
+			},
+		},
+	}
+
+	modelVersionReq := proto.PutContextTypeRequest{
+		CanAddFields: canAddFields,
+		ContextType: &proto.ContextType{
+			Name: modelVersionTypeName,
+			Properties: map[string]proto.PropertyType{
+				"description": proto.PropertyType_STRING,
+				"model_name":  proto.PropertyType_STRING,
+				"version":     proto.PropertyType_STRING,
+				"author":      proto.PropertyType_STRING,
+				"state":       proto.PropertyType_STRING,
+			},
+		},
+	}
+
+	modelArtifactReq := proto.PutArtifactTypeRequest{
+		CanAddFields: canAddFields,
+		ArtifactType: &proto.ArtifactType{
+			Name: modelArtifactTypeName,
+			Properties: map[string]proto.PropertyType{
+				"description":          proto.PropertyType_STRING,
+				"model_format_name":    proto.PropertyType_STRING,
+				"model_format_version": proto.PropertyType_STRING,
+				"storage_key":          proto.PropertyType_STRING,
+				"storage_path":         proto.PropertyType_STRING,
+				"service_account_name": proto.PropertyType_STRING,
+			},
+		},
+	}
+
+	servingEnvironmentReq := proto.PutContextTypeRequest{
+		CanAddFields: canAddFields,
+		ContextType: &proto.ContextType{
+			Name: servingEnvironmentTypeName,
+			Properties: map[string]proto.PropertyType{
+				"description": proto.PropertyType_STRING,
+			},
+		},
+	}
+
+	inferenceServiceReq := proto.PutContextTypeRequest{
+		CanAddFields: canAddFields,
+		ContextType: &proto.ContextType{
+			Name: inferenceServiceTypeName,
+			Properties: map[string]proto.PropertyType{
+				"description":         proto.PropertyType_STRING,
+				"model_version_id":    proto.PropertyType_INT,
+				"registered_model_id": proto.PropertyType_INT,
+				// same information tracked using ParentContext association
+				"serving_environment_id": proto.PropertyType_INT,
+				"runtime":                proto.PropertyType_STRING,
+				"desired_state":          proto.PropertyType_STRING,
+			},
+		},
+	}
+
+	serveModelReq := proto.PutExecutionTypeRequest{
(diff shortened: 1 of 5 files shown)


lampajr (dismissed):
LGTM, thanks a lot @tarilabs !

lampajr:
Just for the records, in a followup PR we could move some tests from `core_test.go` to a dedicated `mlmdtypes_test.go` as they are actually testing the MLMD types setup, e.g., these ones https://github.com/opendatahub-io/model-registry/blob/2cebe8f6d2f0d876b9e80f4d506277bb99a1c54f/pkg/core/core_test.go#L244-L533 where the main purpose is to ensure the types are correctly setup.

I would suggest to move it in a separate PR (if we agree on the change itself obv) as this is not a urgent topic right now and it could require some refactoring in the testcontainers setup.

tarilabs:
...or we could simply rename core_test.go into integration_test.go since it's what's actually is 🙃

lampajr:
> ...or we could simply rename core_test.go into integration_test.go since it's what's actually is 🙃

fair enough, valid alternative :smile:

internal/mlmdtypes/mlmdtypes.go
```diff
@@ -0,0 +1,144 @@
...
+
+var (
+	registeredModelTypeName    = apiutils.Of(constants.RegisteredModelTypeName)
+	modelVersionTypeName       = apiutils.Of(constants.ModelVersionTypeName)
+	modelArtifactTypeName      = apiutils.Of(constants.ModelArtifactTypeName)
+	servingEnvironmentTypeName = apiutils.Of(constants.ServingEnvironmentTypeName)
+	inferenceServiceTypeName   = apiutils.Of(constants.InferenceServiceTypeName)
+	serveModelTypeName         = apiutils.Of(constants.ServeModelTypeName)
+	canAddFields               = apiutils.Of(true)
+)
```
antisaling, line 21: sorry, but I don't get why we have the `apiutils.Of` function, isn't it easier to simply have

```suggestion
var (
	registeredModelTypeName    = &constants.RegisteredModelTypeName
	modelVersionTypeName       = &constants.ModelVersionTypeName
	modelArtifactTypeName      = &constants.ModelArtifactTypeName
	servingEnvironmentTypeName = &constants.ServingEnvironmentTypeName
	inferenceServiceTypeName   = &constants.InferenceServiceTypeName
	serveModelTypeName         = &constants.ServeModelTypeName
	canAddFields               = &true
)
```

?

lampajr: Unfortunately you cannot dereference a constant or literal, that's why we are used to create that function to avoid having to create a variable just to get the address from it

antisaling: I see, thanks for the explanation!

kamuridesu:
Hello! Could you please answer a question for me? Why use pointers for constants, as in https://github.com/opendatahub-io/model-registry/pull/267/files#diff-04387a737874b37f9edc50aec64cd72c86d4e5e876deae7bee3463a2681ac79cR14?
I'm trying to study the code for learning

tarilabs:
> Hello! Could you please answer a question for me? Why use pointers for constants, as in https://github.com/opendatahub-io/model-registry/pull/267/files#diff-04387a737874b37f9edc50aec64cd72c86d4e5e876deae7bee3463a2681ac79cR14? I'm trying to study the code for learning

Hi @kamuridesu it's more about the receivers of those pointers, the receivers need a pointer (to a string, to a boolean, etc).

On other hand, the pointer of what we need, can be built from a constant; so instead of redefining a brand new pointer each time, we just simplify at the top of the file.

Please notice our need is for those pointer to point to a value which _semantically_ is a constant for us (humans). In Go there is no support afaik of getting address of constants, so we need to assign the constant value to a variable, and take that variable address. Instead of doing that over and over, we just "condensed" this way.

Hope this clarifies?

kamuridesu:
Thank you

pkg/core/core.go
```diff
@@ -36,137 +35,81 @@ type ModelRegistryService struct {
...
-		CanAddFields: canAddFields,
-		ContextType: &proto.ContextType{
-			Name: registeredModelTypeName,
-			Properties: map[string]proto.PropertyType{
-				"description": proto.PropertyType_STRING,
-				"state":       proto.PropertyType_STRING,
-			},
-		},
-	}
+func GetTypeIdsIntoTypesMap(cc grpc.ClientConnInterface) (map[string]int64, error) {
```
antisaling: Nit: `into` usually indicates a transformation. I assume you meant to `get type ids in the types map` here, but the meaning is not the same. To avoid the `the` article, I'd rewrite it as

```suggestion
func GetTypeIdsFromTypesMap(cc grpc.ClientConnInterface) (map[string]int64, error) {
```

tarilabs: The semantic is not of `From` tho?
`FromTypesMap` looks wrong, I'm not procuding something _from a TypesMap_ ?

tarilabs: See https://github.com/opendatahub-io/model-registry/pull/267/commits/5f6504207ce09ad06bd18938199834a133109957 if that solves your nit request please.
I'd really like to move fwd with this PR as it is needed from MC

lampajr (approved):
The failure is unrelated to the change as it fails downloading the `openapi-generator-cli`:
```
#20 [builder 14/14] RUN CGO_ENABLED=1 GOOS=linux GOARCH=amd64 make clean model-registry
#20 0.195 rm -Rf ./model-registry internal/ml_metadata/proto/*.go internal/converter/generated/*.go pkg/openapi
#20 0.202 protoc -I./api/grpc --go_out=./internal --go_opt=paths=source_relative \
#20 0.202 	--go-grpc_out=./internal --go-grpc_opt=paths=source_relative api/grpc/ml_metadata/proto/metadata_store.proto
#20 0.282 protoc -I./api/grpc --go_out=./internal --go_opt=paths=source_relative \
#20 0.282 	--go-grpc_out=./internal --go-grpc_opt=paths=source_relative api/grpc/ml_metadata/proto/metadata_store_service.proto
#20 0.427 openapi-generator-cli validate -i api/openapi/model-registry.yaml
#20 0.496 /workspace/bin/openapi-generator-installation/lib/node_modules/@openapitools/openapi-generator-cli/node_modules/@nestjs/common/file-stream/streamable-file.js:36
#20 0.496             this.options.length ??= bufferOrReadStream.length;
#20 0.496                                 ^^^
#20 0.496 
#20 0.496 SyntaxError: Unexpected token '??='
#20 0.496     at wrapSafe (internal/modules/cjs/loader.js:1029:16)
#20 0.496     at Module._compile (internal/modules/cjs/loader.js:1078:27)
#20 0.496     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1143:10)
#20 0.496     at Module.load (internal/modules/cjs/loader.js:979:32)
#20 0.496     at Function.Module._load (internal/modules/cjs/loader.js:819:12)
#20 0.496     at Module.require (internal/modules/cjs/loader.js:1003:19)
#20 0.496     at require (internal/modules/cjs/helpers.js:107:18)
#20 0.496     at Object.<anonymous> (/workspace/bin/openapi-generator-installation/lib/node_modules/@openapitools/openapi-generator-cli/node_modules/@nestjs/common/file-stream/index.js:4:22)
#20 0.496     at Module._compile (internal/modules/cjs/loader.js:1114:14)
#20 0.496     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1143:10)
#20 0.499 make: *** [Makefile:70: openapi/validate] Error 1
#20 ERROR: process "/bin/sh -c CGO_ENABLED=1 GOOS=linux GOARCH=amd64 make clean model-registry" did not complete successfully: exit code: 2
```

tarilabs:
> The failure is unrelated to the change as it fails downloading the `openapi-generator-cli`

yep, also taking note in: https://github.com/opendatahub-io/model-registry/pull/270#pullrequestreview-1838383031

tarilabs:
This has been open for 9 days now and I didn't notice any substantial request for changes. All green. Merging and if any change required can be dealt with in subsequent PRs I guess.

pkg/core/core.go
antisaling: Thanks for addressing it properly while I was away. Regarding the confusion, the new name is definitely clearer, well done :)

PR #272, opened 2024-01-24
update to python/client README to give commands for running python te…

tonyxrmdavidson:
Update to give clear intstructions on how to run tests on a mac m2 machine


clients/python/README.md
```diff
@@ -78,4 +78,28 @@ Common tasks, such as building documentation and running tests, can be executed
...
+cd clients/python
+```
+
+```
+poetry add "https://github.com/tarilabs/ml-metadata/releases/download/1.14.0/ml_metadata-1.14.0-py3-none-any.whl"
+```
+*note: when you add this python whl it will update both your pyprojec.toml and poetry.lock files. **DO NOT COMMIT** these changes.*
+```
+poetry install
+```
```
tarilabs: I would prefer to guide Dev users to alternatives provided in TOML file already:

https://github.com/opendatahub-io/model-registry/blob/0497cf4b325e509e8adabaf45f3fbf3864858d80/clients/python/pyproject.toml#L13-L16

and instruct them to select alternatives by uncommenting/commenting.

Currently working to migrate some of those releases to ODH org at least.

clients/python/README.md
```diff
@@ -78,4 +78,28 @@ Common tasks, such as building documentation and running tests, can be executed
 
 Use `nox -l` to list sessions and execute them using `nox -s [session]`.
 
+## Running Locally on Mac M1 or M2 (arm64 architecture)
+
+If you want run tests locally you will need to set up a colima develeopment environment using the instructions [here](https://github.com/opendatahub-io/model-registry/blob/main/CONTRIBUTING.md#colima)
```
antisaling, line 83: Nit: typo

```suggestion
If you want run tests locally you will need to set up a colima development environment using the instructions [here](https://github.com/opendatahub-io/model-registry/blob/main/CONTRIBUTING.md#colima)
```

clients/python/README.md
```diff
@@ -78,4 +78,28 @@ Common tasks, such as building documentation and running tests, can be executed
 
 Use `nox -l` to list sessions and execute them using `nox -s [session]`.
 
+## Running Locally on Mac M1 or M2 (arm64 architecture)
```
antisaling: This should be a subsubsection, under _development_.

clients/python/README.md
```diff
@@ -78,4 +78,28 @@ Common tasks, such as building documentation and running tests, can be executed
...
+You will also have to change the ml-metadata/..python.whl to one compatible with arm64 architecture. The following command will do this for you.
+
+```
+cd clients/python
+```
+
+```
+poetry add "https://github.com/tarilabs/ml-metadata/releases/download/1.14.0/ml_metadata-1.14.0-py3-none-any.whl"
+```
+*note: when you add this python whl it will update both your pyprojec.toml and poetry.lock files. **DO NOT COMMIT** these changes.*
```
antisaling: While this is correct in some sense, I'd rather say that the user will have to change the *package source* to one compatible with ARM64, and then you can just reference them to the lines commented on the `pyproject.toml` file, as noted by Matteo.

I'd also get rid of this note, the users can do whatever they want in their forks.

Note: if you want to leave instructions for guidance, you can put them all together in a single code block annotated as `sh`. You can also omit the `cd` command IMO.

isinyaaa: Note: if you want to leave instructions for guidance, you can put them all together in a single code block annotated as `sh`.

clients/python/README.md
antisaling: Matteo, could you successfully use `nox` on the Colima/devcontainers setup? If so, that's also what I'd recommend here.

antisaling (changes requested):
Requests inline.

clients/python/README.md
tarilabs: Isn't nox used for testing matrix? 🤔 
I'm not sure I fully understood how this would solve for development environment; happy to try commands/setup, but here I'm afraid I'm not yet following

tonyxrmdavidson: Re tested the process and updated the PR with changes to take take account of comments here.

clients/python/README.md
tonyxrmdavidson: Agree with comments and updated the PR with appropriate changes

clients/python/README.md
tonyxrmdavidson: Moved it under the development heading

antisaling (approved):

tarilabs (approved):

PR #308, opened 2024-02-14
Automate openapi server generation

### LINKED ISSUE #309: Automate OpenAPI Server Generation
Right now the OpenAPI server generation is not automatically triggered therefore we could miss some changes while updating the openapi spec.

Include the openapi server re-generation into the `gen` target.
Make into conditional, such that it is really re-generated if and only if there are changes in the openapi spec.

n/a

**Additional context**
n/a

lampajr:
Fixes https://github.com/opendatahub-io/model-registry/issues/309

In addition it fixes the `type_assert.patch` which was _corrupted_ by the pre-commit hook

## Description

Fixed the patch by restoring required blank lines and spaces.

Auto generate `AssertDocArtifact*` methods by running `make gen/openapi-server`.

## How Has This Been Tested?

Running `make clean gen gen/openapi-server` you shouldn't see any error and no uncomitted changes

## Merge criteria:


tarilabs (dismissed):
nice catch! thank you so much 🙏 is there a way we can ensure not to miss this in the future in your opinion in an easy way? 🤔 

atm I can only think of just programmatically do it, but I wonder if it's too much work... maybe we should put some comments in the ModelArtifact, DocArtifact that due to the way the codegen works, this method is to be expected?

lampajr:
> is there a way we can ensure not to miss this in the future in your opinion in an easy way?

Well for the type_assert.patch issue I would remove it from the pre-commit formatting if possible, @isinyaaa is it possible to do that?

Regarding the `Assert*` methods, actually they are not used in the server (at least for some of them that I checked).. therefore we could also evaluate if we need that file in the future but yeah I think the best practice would be to always run `make gen/openapi-server` whenever we change the openapi spec (right now this is not automatically triggered but we could think to include it as well as we do for the client/model part)

tarilabs:
> right now this is not automatically triggered

yes, this part.

tarilabs:
> I would remove it from the pre-commit formatting if possible

try with adding ` exclude: '^patches/'` just at the very root of the `.pre-commit-config.yaml`.

(edit: I was not able to configure per-hook)

.pre-commit-config.yaml
```diff
@@ -16,6 +16,7 @@ repos:
       - id: detect-private-key
       - id: end-of-file-fixer
       - id: trailing-whitespace
+    exclude: '^patches/'
```
tarilabs: did this work for you? I got a "wrong format" error message of sort? 🤔

lampajr: Sorry I did not try that yet, I wrongly included that in the commit.. trying right now

lampajr: As you suggested I tried adding that line at root of the file and it seems working.

tarilabs: > As you suggested I tried adding that line at root of the file and it seems working.

🥳 🚀

tarilabs: we can use analogously if the linter is going _beyond_ the Go ootb Linter hence colliding with the .patch, in case

lampajr: +1 I agree

lampajr:
Ok I was able to automate the openapi-server generation but there are some issues in the `build` workflow, somehow the build generates some uncomitted changes, instead I tried running the same command multiple times locally and I got the same result.

So for now I think we can just remember to run `make gen/openapi-server` whenever we made changes to the api spec

antisaling (dismissed):
Thanks a lot for fixing that!

lampajr:
As suggedted by @tarilabs I was able to automate the openapi server generation only if there are some changes (even if staged) in the `api/openapi/model-registry.yaml` file.

This way the server generation won't happen during the GHA.

lampajr:
Additional comment is that you can always force the re-generation by setting `FORCE_SERVER_GENERATION` env variable, e.g., by running `make FORCE_SERVER_GENERATION=true gen/openapi-server`

tarilabs (approved):
Awesome, thank you so much, and glad some of the suggestions were relevant! 🚀 👍

PR #315, opened 2024-02-19
This commit will add nehachopra27 to the OWNERS file.

tonyxrmdavidson:
This commit will add nehachopra27 to the OWNERS file.  This is needed as Neha will be working on the openshift-ci deployment.


tarilabs (approved):
/lgtm but @tonyxrmdavidson seems to me you need first `nehachopra27` in the org? 🤔

